PR #3, opened 2018-03-20
Add TravisCI config and README badges

mmasi-attivio:
- Add TravisCI configuration
- Modifiy the pre-existing generic badges to point to the real URLs

Note that if this PR is accepted, it should _**not**_ be merged into master as-is -- the urls are set to point to my fork of the repo so it won't reflect reality. This is more of a POC to show what the badges will look like.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,7 @@
+language: node_js
+node_js:
+  - '8.9.4'
+install:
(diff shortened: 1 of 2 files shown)


mmasi-attivio:
Thanks Luis!  Changes have been merged into master.

PR #4, opened 2018-04-03
Issue-5: Use `displayName` instead of class `name` for setting default properties

voutilad:
JS prod minimizers (like UglifyJS) will mangle the class names, so using the `name` attribute is dangerous. Currently, searchui-frontend breaks and can't use certain properties (i.e. `businessCenterProfile`). This change (and future addition of `displayName` static attributes, allows proper production optimizations.

Other components wrapped by `Configurable` should get `displayName` attributes in the future as it also enhances React debugability in production.


lvaldezattivio:
The code in Configurable.js has been merged and every component now defines a displayName property. The validation script that will be added shortly also checks to ensure that every future component has this property as well.

PR #12, opened 2018-05-11
Feature/plat 41511

lvaldezattivio:


src/api/Signals.js
```diff
@@ -26,44 +27,13 @@ export default class Signals {
...
-              `${response.statusText} (error code ${response.status})` :
-              `Unknown error of type ${response.status}`;
-            console.log('Failed to submit signal', updatedSignal, message);
-          }
-        },
-        (error: any) => {
-          // Catch network-type errors from the main fetch() call
+      const callback = (response: any | null, error: string | null) => {
+        if (!error) {
           console.log('Failed to submit signal', updatedSignal, error);
```
lvaldezattivio: This seems to be logging the failure if there's no error... is that what we want?

src/util/AuthUtils.js
```diff
@@ -232,38 +233,21 @@ export default class AuthUtils {
...
+          console.log('Got an error parsing the current user info', error);
+        }
+        callback(userInfo);
+      };
+      FetchUtils.fetch(
+        `${AuthUtils.config.ALL.baseUri}/rest/serverDetailsApi/user`,
+        null,
+        fetchResponseCallback,
+        'GET',
+        'Got an error parsing the current user info');
```
lvaldezattivio: This should be more generic, since it could be parsing but it could also be fetching... Maybe "Got an error retrieving the current user\u2019s details." (Note that the error messages should generally have a period at the end since they're full sentences... also note the \u2019 for the "real" apostrophe ’ as opposed to the fake one '.)

src/components/AutoCompleteInput.js
```diff
@@ -115,22 +118,16 @@ export default class AutoCompleteInput extends React.Component<AutoCompleteInput
...
-          this.setState({
-            isLoading: false,
-            suggestions: [],
-            error,
-            open: error.length > 0,
-          });
-        });
+          }
+        };
+        FetchUtils.fetch(`${uri}?term=${encodedValue}`, null, callback, 'GET', 'An error occured while looking for suggestions: ');
```
lvaldezattivio: Since the error message might occur as-is, I'd change it to end with a period rather than the colon, and no trailing space since the FetchUtils class adds that.

src/api/Signals.js
Akshay-Choudhary: This seems to be incorrect. We should log the errors if there are any. I would remove the negation in that case.

PR #16, opened 2018-05-23
PLAT-41671: Added properties to support legend alignment and height adjustment to the pie chart.

Akshay-Choudhary:


src/components/PieChartFacetContents.js
```diff
@@ -17,10 +17,21 @@ type PieChartFacetContentsProps = {
...
    */
   entityColors: Map<string, string>;
+  /** Property to align the legend to left, right or center. */
+  legendAlign?: 'left' | 'right' | 'center';
+  /** Property to position the legend either as horizontal or vertical. */
+  legendLayout?: 'horizontal' | 'vertical';
+  /** Property to align the legend vertically either on top, middle or bottom. */
+  legendVerticalAlign?: 'top' | 'middle' | 'bottom';
+  /** Height of the Pie Chart. */
+  height?: number;
```
lvaldezattivio: You show the type of height as number but for the default value you say "400px," which is a string. The type should be string, I think.

src/components/PieChartFacetContents.js
```diff
@@ -17,10 +17,21 @@ type PieChartFacetContentsProps = {
...
 };
 
 type PieChartFacetContentsDefaultProps = {
   entityColors: Map<string, string>;
+  legendAlign?: 'left' | 'right' | 'center';
+  legendLayout?: 'horizontal' | 'vertical';
+  legendVerticalAlign?: 'top' | 'middle' | 'bottom';
```
lvaldezattivio: You don't list the height property in the PieChartFacetContentsDefaultProps type definition.

src/components/PieChartFacetContents.js
```diff
@@ -17,10 +17,21 @@ type PieChartFacetContentsProps = {
    * the map are ignored.
    */
   entityColors: Map<string, string>;
+  /** Property to align the legend to left, right or center. */
```
lvaldezattivio: The comments for properties that have default values should indicate what the default values are. So the user knows what will happen if they omit the property.

src/components/PieChartFacetContents.js
```diff
@@ -17,10 +17,21 @@ type PieChartFacetContentsProps = {
...
+  legendAlign?: 'left' | 'right' | 'center';
+  /** Property to position the legend either as horizontal or vertical. */
+  legendLayout?: 'horizontal' | 'vertical';
```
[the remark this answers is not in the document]
Akshay-Choudhary: legendAlign represents the horizontal alignment of legend relative to the chart while legendLayout represents how the individual elements in the legend should appear(Horizontally or Vetically). I should add more specific comments comparing what the highcharts documentation says.

src/components/PieChartFacetContents.js
Akshay-Choudhary: Should I include the option values as well in the comments apart from default values? For e.g "Property representing horizontal alignment of the legend box within the chart area. Possible values are left, right or center. Default value is center." Or I should omit the Possible values part and just keep the default values since we are mentioning in the actual property declaration what the possible values are...

lvaldezattivio: Yes, that sounds good with the options included.

src/components/PieChartFacetContents.js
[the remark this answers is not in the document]
lvaldezattivio: Okay.

Akshay-Choudhary:
Checked in all the changes requested as per this pull request. Let me know if anything else needs to be changed.

PR #19, opened 2018-05-31
PLAT-41784 DefaultImage component

benmvachon:


src/components/DefaultImage.js
```diff
@@ -0,0 +1,60 @@
...
+import React from 'react';
+
+type DefaultImageProps = {
+  /**
+   * The address of the image you want to display.
+   */
+  imageSrc: string;
+  /**
+   * The address of a default image to display if there's an issue displaying the prefered image address.
+   */
```
lvaldezattivio: Properties which have a default value shouldn't be set to have an optional ("?") type. They always, in fact, have a value because of the fact that they are defined in the XXXDefaultProps type and are initialized in the defaultProps object. (Also, the type of a property in the XXXDefaultProps type should always be identical to its type in the XXXProps type.)

src/components/DefaultImage.js
```diff
@@ -0,0 +1,60 @@
...
+
+type DefaultImageDefaultProps = {
+  defaultSrc: string | null;
+}
+
+type DefaultImageState = {
+  src: string | null;
+};
+
+/**
```
lvaldezattivio: I would include some of what the original component's comment says about the img properties, etc. The original DisappearingImage component's comment is here:
`Component that will hide an image if it can't be displayed due to, e.g.,
a URL to a non-existant file. You can pass in any props that work for
a standard  tag and they'll be added to the inserted image.`

src/components/DefaultImage.js
```diff
@@ -0,0 +1,60 @@
+// @flow
+
+import React from 'react';
+
+type DefaultImageProps = {
+  /**
+   * The address of the image you want to display.
+   */
+  imageSrc: string;
+  /**
```
lvaldezattivio: In the comment for the defaultSrc property, I would mention that it's optional. The styleguidist stuff doesn't seem to do the best job with identification of default properties in the Rest components.

PR #22, opened 2018-06-29
Feature/plat 41673

Akshay-Choudhary:


src/util/AuthUtils.js
```diff
@@ -19,13 +19,13 @@ export default class AuthUtils {
...
       throw configError;
     }
-    if (config.ALL.authType === 'XML') {
-      // Only validate users if the auth type is XML
-      const usersError = AuthUtils.validateUsers(users);
-      if (usersError) {
-        throw usersError;
-      }
-    }
+    // if (config.ALL.authType === 'XML') {
```
lvaldezattivio: Why are we not doing this validation any longer?

src/util/AuthUtils.js
```diff
@@ -172,11 +172,14 @@ export default class AuthUtils {
 
   /**
    * Check whether the user has a particular permission.
-   * TO BE IMPLEMENTED
    */
-  static hasPermission(/* user: any, permission: string */): boolean {
+  static hasPermission(user: any, permission: string): boolean {
     if (AuthUtils.config && AuthUtils.config.ALL && AuthUtils.config.ALL.authType === 'NONE') {
```
lvaldezattivio: Why are we not doing the check when the authType is set to "NONE"? This is the case when Search UI is being run as a module on the node instead of with the servlet.

src/util/AuthUtils.js
```diff
@@ -229,7 +232,7 @@ export default class AuthUtils {
     const userObject = AuthUtils.getSavedUser();
     if (userObject && userObject.timeout && userObject.timeout > new Date().getTime()) {
       callback(userObject);
-    } else if (AuthUtils.config.ALL.authType === 'SAML' || AuthUtils.config.ALL.authType === 'NONE') {
+    } else if (AuthUtils.config.ALL.authType === 'SAML' || AuthUtils.config.ALL.authType === 'NONE' || AuthUtils.config.ALL.authType === 'XML') {
```
lvaldezattivio: Why is the XML case added here? If we're using XML authentication, then there is no callback to use to get the user... we will already have it.

Akshay-Choudhary:
- Just tested the changes we made yesterday in our meeting and made sure that role based authentication works. I added a user in the Connector user role and one in Connector Admin role.
- In the admin ui frontend code added a condition to show only the connectors pane when a user is part of the Connector user role and it worked as expected.(This was just a dummy test to check if the AuthUtils code for hasPermission works)

PR #28, opened 2018-07-23
issue-27: Add Facet Search

brandon-bogan:
Creates a new component called `FacetSearchBar`, which can wrap a facet component (e.x. `MoreListFacetContents`), with the ability to display a Search Bar above the facet list, and an export button below. 

The Search Bar allows users to search for values outside of the top n displayed in the list (like if they want to filter to news showing a specific company, but that company isn't in the top buckets shown in the list). They can click the search result for the facet to apply it as a filter. 

The export button allows users to download a CSV containing a list of all facet values for that facet, and the count of docs that match that facet value. 

Here is an exmaple screenshot:
example

The `Facet` component was also modified, to wrap the MoreListFacetContents component with the FacetSearchBar component. 

Note that users can configure the `FacetSearchBar`. Here is a sample configuration that can be placed in configuration.properties.js:
```javascript
 FacetSearchBar: {
	// SEARCH BAR CONGIGS
	// Should there be a Facet Search Bar exposed to end users 
	showSearchBar: true,
	// The placeholder for the search bar
        placeholder: 'Search Facet Values',
	// The label on the 'Search' button
        buttonLabel: 'Search',
	// Max number of matching facet values to show
	maxValues: 5,
	// EXPORT CONFIGS
	// Should there be an export button exposed to end users
	showExportButton: true,
	// The label for the export button for exporting results to a CSV
	exportButtonLabel: 'Export',
  },

### DIFF
--- a/src/components/Facet.js
+++ b/src/components/Facet.js
@@ -42,12 +42,6 @@ type FacetProps = {
   bordered: boolean;
   /** Controls the colors used to show various entity types (the value can be any valid CSS color) */
   entityColors: Map<string, string>;
-  /** Should the search bar be exposed for list facets */
-  showSearchBar: boolean;
-  /** Should the export button be available for the facet values for list facets */
-  showExportButton: boolean;
-  /** The label for the export button */
-  exportButtonLabel: string;
 }
 
 type FacetDefaultProps = {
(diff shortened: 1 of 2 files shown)


lvaldezattivio:
Merged from a different branch with some tweaks. There are still a known issue: clicking outside the pop-up with the search results doesn't close it.

PR #34, opened 2018-08-06
issue-34: Use Color Map with fields for Pie Chart

nikhilj13:
Modifies Facet component to take in an optional prop fieldValueColors which can be specified in the configuration.properties.js file ([in SearchUI](https://github.com/attivio/searchui/blob/master/frontend/configuration.properties.js)). If a sector is selected, the pie chart will update to the color of that sector instead of using the first value in entityColors prop. Here's an example:
color-map-fix
Users can configure the colors used for any fields that are used for creating pie charts. Here's an example configuration placed in configuration.properties.js for the above pie chart which uses the field table:
```
fieldValueColors: {
		table: {
			'Employees': '#007dbc',
			'Email': '#ed7a23',
			'News': '#fedd0e',
			'Research': '#db2e75',
			'Grant Proposal': '#ef8baa',
			'Dissertation': '#fcb62c',
		},
		sentiment: {
			pos: 'lightgreen',
			neg: '#ff6666',
		},
	},
```

### DIFF
--- a/src/components/Facet.js
+++ b/src/components/Facet.js
@@ -43,7 +43,7 @@ type FacetProps = {
   /** Controls the colors used to show various entity types (the value can be any valid CSS color) */
   entityColors: Map<string, string>;
   /** Controls the colors (specified in configurations) for specific fields used in pie charts */
-  fieldValueColors?: Map<string, Map<string, string>>;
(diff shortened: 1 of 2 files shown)


klk1010:
originating repo has been deleted. @nikhilj13 will redo PR

PR #36, opened 2018-08-07
issue-35: Optional callback for SearchBar component

nikhilj13:
Modifies the Searchbar component to be able to take in a callback function as an optional prop. This can be particularly useful in cases when users have to perform one or more custom queries based on the value in search bar, and in situations when simply relying on the results from searcher is not sufficient.

### DIFF
--- a/src/components/SearchBar.js
+++ b/src/components/SearchBar.js
@@ -182,16 +182,14 @@ class SearchBar extends React.Component<SearchBarDefaultProps, SearchBarProps, S
     const searcher = this.context.searcher;
     if (searcher) {
       if (doSearch) {
+        searcher.updateQuery(newQuery);
         // Use the searchFunction callback if it is passed in as a prop
         if (this.props.searchFunction) {
-          searcher.updateQuery(newQuery);
           this.props.searchFunction();
         } else {
           searcher.setQueryAndSearch(newQuery);
           this.route();
         }
-      } else {
-        searcher.updateQuery(newQuery);
       }
     }
     this.forceUpdate();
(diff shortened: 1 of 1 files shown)


src/components/SearchBar.js
```diff
@@ -180,10 +182,14 @@ class SearchBar extends React.Component<SearchBarDefaultProps, SearchBarProps, S
     const searcher = this.context.searcher;
     if (searcher) {
       if (doSearch) {
-        searcher.setQueryAndSearch(newQuery);
-        this.route();
-      } else {
         searcher.updateQuery(newQuery);
+        // Use the searchFunction callback if it is passed in as a prop
+        if (this.props.searchFunction) {
```
luigivaldez: So, here, we'd need to do something like:
```
let doSearch = true;
if (this.props.searchFunction) {
  doSearch = this.props.searchFunction(newQuery);
}
if (doSearch) {
  searcher.setQueryAndSearch(newQuery);
  this.route();
}
```

Does that make sense?

src/components/SearchBar.js
```diff
@@ -213,7 +219,9 @@ class SearchBar extends React.Component<SearchBarDefaultProps, SearchBarProps, S
 
   doSearch() {
     const searcher = this.context.searcher;
-    if (this.props.route && searcher) {
+    if (this.props.searchFunction) {
```
luigivaldez: This would need to be modified too, to do something similar to what happens in the `updateQuery()` method above.

luigivaldez (changes requested):
Some thoughts on this callback....

luigivaldez:
PR 177 replaces this.

PR #44, opened 2018-09-21
issue-43: MiniSearchUI can now be integrated with "external" results and components

brandon-bogan:
MiniSearchUI can now take optional props for results to render. These props, in one way or another, are propagated on to SearchResultsCount and SearchResults, and so these components were also updated as part of this new feature. 

MiniSearchUI also takes props for  functions to call when the user enters a query and/or clicks search, so that the results being passed in can be updated accordingly.

### DIFF
--- a/src/components/MiniSearchUI.js
+++ b/src/components/MiniSearchUI.js
@@ -8,6 +8,7 @@ import SearchResults from '../components/SearchResults';
 import Scrollable from '../components/Scrollable';
 import SearchResultsCount from '../components/SearchResultsCount';
 import ObjectUtils from '../util/ObjectUtils';
+import QueryResponse from '../api/QueryResponse';
 
 type MiniSearchUIProps = {
   /**
@@ -15,11 +16,19 @@ type MiniSearchUIProps = {
    * the MiniSearchUI. Optional—defaults to 100%.
    */
   scale: number;
+  onSearch?: (q: string) => void;
+  updateQuery?: (q: string) => void;
+  response?: QueryResponse;
+  error?: string | null;
 };
 
 type MiniSearchUIDefaultProps = {
   scale: number;
 };
+
+type MiniSearchUIState = {
+  query: string;
+}
 
 /**
  * A miniature, self-contained component that presentds super simple search UI including a text field for the
@@ -27,7 +36,7 @@ type MiniSearchUIDefaultProps = {
  * showing the resulting documents. It must be nested inside a Searcher component and will use that parent
  * Searcher to manage its state.
  */
-export default class MiniSearchUI extends React.Component<MiniSearchUIDefaultProps, MiniSearchUIProps, void> {
+export default class MiniSearchUI extends React.Component<MiniSearchUIDefaultProps, MiniSearchUIProps, MiniSearchUIState> {
   static defaultProps = {
     scale: 1.0,
   };
@@ -40,18 +49,30 @@ export default class MiniSearchUI extends React.Component<MiniSearchUIDefaultPro
 
   constructor(props: MiniSearchUIProps) {
     super(props);
+    this.state = {
+      query: '*',
+    };
     (this: any).doSearch = this.doSearch.bind(this);
     (this: any).updateSearchQuery = this.updateSearchQuery.bind(this);
   }
 
   doSearch() {
-    const searcher = this.context.searcher;
-    searcher.doSearch();
+    if (this.props.onSearch) {
+      this.props.onSearch(this.state.query);
+    } else {
+      const searcher = this.context.searcher;
+      searcher.doSearch();
+    }
   }
 
   updateSearchQuery(query: string) {
-    const searcher = this.context.searcher;
-    searcher.updateQuery(query);
+    if (this.props.updateQuery) {
+      this.props.updateQuery(query);
+    } else {
+      const searcher = this.context.searcher;
+      searcher.updateQuery(query);
+    }
+    this.setState({ query });
   }
 
   render() {
@@ -60,12 +81,16 @@ export default class MiniSearchUI extends React.Component<MiniSearchUIDefaultPro
         <NavbarSearch
           onSearch={this.doSearch}
           updateSearchString={this.updateSearchQuery}
-          value={this.context.searcher.state.query}
+          value={this.state.query}
           style={{
             marginLeft: '8px',
           }}
         />
-        <SearchResultsCount style={{ marginLeft: '20px', paddingBottom: '8px' }} />
+        <SearchResultsCount
+          style={{ marginLeft: '20px', paddingBottom: '8px' }}
+          response={this.props.response}
+          error={this.props.error}
+        />
         <Scrollable
           style={{
             height: '428px',
@@ -81,6 +106,7 @@ export default class MiniSearchUI extends React.Component<MiniSearchUIDefaultPro
             style={{
               transform: `scale(${this.props.scale}, ${this.props.scale})`,
             }}
+            response={this.props.response}
           />
         </Scrollable>
       </div>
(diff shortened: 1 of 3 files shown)


src/components/MiniSearchUI.js
```diff
@@ -7,28 +7,62 @@ import NavbarSearch from '../components/NavbarSearch';
...
   /**
    * A scale (1.0 = 100%) to use when rendering the search results part of
    * the MiniSearchUI. Optional—defaults to 100%.
    */
   scale: number;
+  /**
+   * Optional callback to handle searching if not using the context's
+   * Searcher component to do the searching (otherwise the searcher's onSearch method is used).
+   */
+  onSearch: null | (q: string) => void;
```
klk1010: Indicate this is optional: `onSearch?: (q: string) => void;`

The difference between 
A: `onSearch: null | (q: string) => void;` and
B: `onSearch?: (q: string) => void;`

In example A, you are specifying that the user must explicitly pass a prop named `onSearch` and it may have the value of `null` or it could be a function like `(q: string) => void`. Omitting a function is not the same as `null`. If `onSearch` was not passed as a prop, it would be `undefined`. It's easier to quickly see that something is optional using `?` - that's it's specific intention.

src/components/MiniSearchUI.js
```diff
@@ -7,28 +7,62 @@ import NavbarSearch from '../components/NavbarSearch';
...
+  /**
+   * Optional query error to render if not using the context's
+   * Searcher component to do the searching (otherwise the searcher's state's error is used).
+   */
+  error: string | null;
 };
 
 type MiniSearchUIDefaultProps = {
   scale: number;
+  onSearch: null | (q: string) => void;
```
klk1010: Unlike `MiniSearchUIProps`, `MiniSearchUIDefaultProps` is explicitly defined in your component. defaultProps are based on your specifications on lines `60-66` and are not based on what the parent passes through as a prop. Keeping that in mind, `onSearch`, `updateQuery`, `response`, and `error`, can only be `null` here.

src/components/SearchResults.js
```diff
@@ -63,6 +70,8 @@ type SearchResultsDefaultProps = {
   showScores: boolean;
   showTags: boolean;
   showRatings: boolean;
+  response: QueryResponse | null;
```
klk1010: As a `defaultProp`, `response` in this instance can only be `null`.

src/components/SearchResults.js
```diff
@@ -55,6 +55,13 @@ type SearchResultsProps = {
   showRatings: boolean;
   /** A style to apply to the results list */
   style: any;
+  /**
+   * A response can be passed in from custom searches if you don't want
+   *  to use the response on the searcher in this.context
+   */
+  response: QueryResponse | null;
```
klk1010: This should be made optional for backwards compatibility.
`response?: QueryResponse | null;`

src/components/SearchResults.js
```diff
@@ -55,6 +55,13 @@ type SearchResultsProps = {
...
+  /** Offset of search results for long scrolling or pagination */
+  offset: number;
```
klk1010: This should be made optional for backwards compatibility.
`offset?: number;`

src/components/SearchResults.js
```diff
@@ -88,8 +99,8 @@ export default class SearchResults extends React.Component<SearchResultsDefaultP
 
   renderResults() {
     const searcher = this.context.searcher;
-    const response = searcher.state.response;
-    const offset = searcher.state.resultsOffset;
+    const response = this.props.response !== null ? this.props.response : searcher.state.response;
+    const offset = this.props.response ? this.props.offset : searcher.state.resultsOffset;
```
klk1010: Make `offset` optional for backwards compatibility.

src/components/SearchResultsCount.js
```diff
@@ -2,50 +2,79 @@
...
-        );
-      } else if (searcher.state.error) {
-        // got an error...
-        message = `Error: ${searcher.state.error}`;
+    if (this.props.response) {
+      response = this.props.response;
+    } else {
+      const searcher = this.context.searcher;
+      if (searcher) {
+        response = searcher.state.response;
```
klk1010: This is a ninja modifier. You are modifying props here which is not immediately obvious, but breaks the Golden Rule.
On `line 45`, you are passing the reference of `this.props.response` to a new variable. Because of that, you are directly updating `this.props.response` on this line.

src/components/SearchResultsCount.js
```diff
@@ -2,50 +2,79 @@
...
+    error: null,
+  }
+
   static contextTypes = {
     searcher: PropTypes.any,
   };
 
   static displayName = 'SearchResultsCount';
 
   render() {
```
klk1010: Here's an example of a more functional approach. It follows immutable principles so you never modify props. It also cleans up the main render by moving the logic out into a new function.
```
renderMessage = () => {
  const { response, error } = this.props;
  const { searcher } = this.context;
  const derivedResponse = response ? response : searcher.state.response | null;

  let message;
  let countMessage;

  if (response) {
    const count = response.totalHits;
    if (count === 0 || typeof count !== 'number') {
      countMessage = 'No results found';
    } else if (count === 1) {
      countMessage = '1 result found';
    } else {
      countMessage = `${count} results found`;
    }
    return (
      <span>
        {countMessage}
        <span style={{ fontWeight: 'normal' }}>
          {' '}
          (in {response.totalTime}ms)
        </span>
      </span>
    );
  } else if (error) {
    return `Error: ${error}`;
  } else if (searcher.state.error) {
      // got an error...
    return `Error: ${searcher.state.error}`;
  } else {
    // Not yet searched...
    return ''; 
  }
}

render() {
  return (
    <div className="attivio-globalmastnavbar-results">
      {this.renderMessage()}
    </div>
  );
}
```

src/components/SearchResultsCount.js
brandon-bogan: I might be missing something, but line 45 is inside an `if` block, and line 49, where you are saying that I'm modifying the props, is in the else case, so line 49 would never be executed if `response` was set equal to props from line 45.

brandon-bogan: Never mind - going with the `renderMessage` approach you suggest below makes this irrelevant anyway and your way is cleaner.

src/components/MiniSearchUI.js
brandon-bogan: Done. Made `onSearch`, `updateQuery`, `response`, and `error` optional props. I also removed the `null` option for `onSearch` and `updateQuery` and removed these props from the default props, since these should either be valid functions or not exist at all as props.

src/components/MiniSearchUI.js
brandon-bogan: Done. Made all defaultProps null, removed `onSearch` and `updateQuery` from defaultProps.

src/components/SearchResults.js
brandon-bogan: Done.

src/components/SearchResults.js
brandon-bogan: Done.

src/components/SearchResults.js
brandon-bogan: Changed to:
`const offset = this.props.response && this.props.offset ? this.props.offset : searcher.state.resultsOffset;`

Is this what you meant?

src/components/SearchResultsCount.js
brandon-bogan: Done. Note that I wrote the 3rd line of `renderMessage` slightly different to avoid a lint error but otherwise incorporated your code suggestion above.

PR #46, opened 2018-09-21
issue-45: Enable Pressing Enter Key to Search

brandon-bogan:
Added an onKeyDown prop to the input component that calls doKeyPress, which checks if the key is Enter or not, and calls onSearch if it is (copied from the implementation of SearchBar)


brandon-bogan:
@luigivaldez - Per your comment, I've copied over the change to `NavbarSearch` in a new branch directly on SUIT and submitted a new PR for that branch. See PR #171 

Closing this PR.

PR #56, opened 2018-10-23
PLAT-42564 no user required for authType 'NONE'

benmvachon:


src/components/AuthRoute.js
```diff
@@ -66,9 +66,13 @@ class AuthRoute extends React.Component<AuthRouteDefaultProps, AuthRouteProps, A
...
 
   render() {
-    // If the user is logged in and has permission for this
-    // route, then just render the route.
-    if (AuthUtils.isLoggedIn(this.props.required)) {
+    // If authentication is required by the app, then make sure the user is logged in.
+    // If there is no general authentication required by the app,
+    // then only check that the user is logged in if specific credentials are required for the page
+    if (
+      (this.props.authType !== 'NONE' && AuthUtils.isLoggedIn(this.props.required)) ||
```
lvaldezattivio: NONE doesn't mean that no authentication is required. It means that the Search UI app isn't doing it (e.g. XML authentication) and the servlet isn't doing it (e.g. SAML authentication). NONE is currently the value that needs to be specified when the user is logged in using the authentication set in the module running in the Attivio node. (The name 'NONE' was a relic of early days in the development when it really did mean there was no authentication; I have been considering changing it to be called 'ATTIVIO' or 'NODE' to make this clearer.)

If 'NONE' is set, then the app still expects to see a logged-in user, obtained by making a REST call to /rest/serverDetailsApi/user. It uses this user when querying Attivio so it can return documents that the user has permission to see. It also uses it to potentially allow access to routes based on the logged-in user (that's what the point of the AuthRoute component is).

If we want to add a new authentication type, I think we should call it something else like 'NO_AUTH'. And we could change the existing 'NONE' to 'NODE' but leave 'NONE' as a synonym for now so that existing users don't start seeing behavior changing.

In this case, the default value in the `configuration.properties.js` file and the documentation comments there would need to be changed as well. (There's a separate ticket—https://jira.attivio.com/browse/RELENG-3302—to have the build process make changes to `configuration.properties.js` based on whether it's destined for the module or the servlet... the text of ticket would need to be updated as well...)

Does this make sense?

benmvachon: That makes sense.
This was done to fix the case where "requires-authentication" is set to false in the deploy-webapp feature for searchui (and rest). Right now, this makes it so that you can access the app, but you'll get a blank white screen.
If auth is "NONE" (a.k.a. "NODE") then the app doesn't need to apply any security, it's all done by the webappserver embedded in the node. So we shouldn't check that there's a user, because one might not even be required.

benmvachon: Although, I think that this will fix the issue in tomcat as well.
When the authType is "NONE", shouldn't it mean that no authentication is required within the searchui app?
If either the node is handling authentication, or the app is served by tomcat and doesn't require a user, then searchui should behave the same way (by not requiring a logged in user).
I don't think we do need a separate authType for when the app is served by the node.

lvaldezattivio: We had said that using Search UI and REST without authentication wasn't a supported use case, so it's technically a feature request and not a bug fix…
That said, I think we should probably get together and layout all of the scenarios and what they should look like before making the changes.

lvaldezattivio: I suppose, including the variation specified by the feature request, there are really three cases:
1) The authentication happens using the users.xml file, right in the JavaScript code of the application and no users are retrieved from the `/rest/serverDetailsApi/user` call. Note that this is not considered secure since the users.xml file is transmitted to and visible inside the browser, and we tell customers it shouldn't be used for production systems—it exists mainly as a way for the Solutions Architects to be able to create quick and dirty PoVs...
2) Authentication happens outside of the JavaScript code of the application, either in the servlet (in which case it's done using SAML) or in the node's Jetty, in which case it's done using whatever authentication provider exists there (and also, importantly, the same authentication provider is used for both the searchui module and the rest module). In this case, the authentication is configured outside of the context of the Search UI application, either in the searchui module's "deploy web app" feature XML for node-hosted instances or in the servlet's `application.properties` file for servlet-hosted instances.
3) [THE NEW SCENARIO] We don't want any the JavaScript code of the application to assume anything about a user being logged in, and just ignore any checks that were requiring that and never even try to call the `/rest/serverDetailsApi/user` endpoint. This will require that the servlet or the module also be configured to not have any authentication enabled, or any REST API calls will fail, so it's not a security hole as anyone could access the REST calls directly in this case anyway. In this case, we need to decide what to do in the application, e.g., maybe to show "No Authentication" or something similar in the masthead where the name of the currently logged-in user is usually displayed.

So, I think it might make sense to set up the options to be:

* "XML" which has the same meaning as today (and applies to either node- or servlet-hosted Search UI),
* "EXTERNAL" which would take the place of today's "SAML" and "NONE" options, since, for all intents and purposes, they're the same as far as the JavaScript code goes, relying on whatever is providing the REST APIs to handle this, and
* "UNSECURED" which would be the new case, again regardless of the way the application is hosted.

How does this sound?

benmvachon: I think I just need an explanation of why "EXTERNAL" and "UNSECURED" need to be distinct.
If authentication is handled outside of the app, why does the app itself need do anything special to require it?
Let's meet at some point to talk about this.

lvaldezattivio: I suppose it is possible that we could update the "user" REST call to return a bogus but filled-in user info record if there is actually nobody logged in since, if the REST API is configured to be authenticated (via SAML in the servlet or via an authentication provider in the node) then the REST API call won't get past the Spring security layer, so if nobody is logged in and the REST call succeeds, then we're in the special "do whatever y'all want" mode. In this scenario, the front end would think someone ("Joe Bogus") was logged in and work "normally," showing the fake user info in in the masthead... (I suppose we ought to use something a little less tongue-in-cheek as the fake name, such as "Unsecured" or even just " "...)

src/components/AuthRoute.js
```diff
@@ -93,6 +87,18 @@ class AuthRoute extends React.Component<AuthRouteDefaultProps, AuthRouteProps, A
...
+    // then only check that the user is logged in if specific credentials are required for our route
+    if (!this.props.required || AuthUtils.isLoggedIn(this.props.required)) {
+      return (
+        <Route
+          {...this.props}
+        />
+      );
+    }
+
+    // insufficient credentials (provided by external authentication system) => don't let the user through
```
lvaldezattivio: Maybe "If no particular authentication is required or one is and the user has it, show the route."

benmvachon:
I think we should leave the routes as `<AuthRoutes>` just so that Solutions doesn't have to do any custom code changes to make XML authentication work.

lvaldezattivio:
Sounds good!

PR #59, opened 2018-10-29
issue-58: Add closeMenu function and RootCloseWrapper

brandon-bogan:
FacetSearchBar will now close its suggestion menu when a user clicks somewhere else


src/components/FacetSearchBar.js
```diff
@@ -236,7 +243,7 @@ class FacetSearchBar extends React.Component<FacetSearchBarDefaultProps, FacetSe
    */
   doSearch() {
     const callback = this.handleSearchResults;
-    this.doConfiguredSearch(`${this.state.facetValue}*`, this.props.maxValues * 2, callback, this.context.searcher);
+    this.doConfiguredSearch(`${this.state.facetValue}*`, this.props.maxValues * 2, callback, this.context.searcher); // `
```
luigivaldez: What's up with this comment? Can you remove it?

luigivaldez (changes requested):
See my comment about your comment.... Otherwise, this looks good.

src/components/FacetSearchBar.js
klk1010:
```suggestion
    this.doConfiguredSearch(`${this.state.facetValue}*`, this.props.maxValues * 2, callback, this.context.searcher);
```

brandon-bogan: Fixed in latest commit

luigivaldez (approved):

PR #63, opened 2018-11-15
Feature/plat 42667

benmvachon:
The new `adminRole` property of these two components will also be configurable in the `configuration.json` file. It might not be a bad idea to eventually add it into that file (in SearchUI) with its default value of `'AIE_Administrator'` (just to make it obvious that it can be configured at this top level).


lvaldezattivio:
Why does this need to be configurable? It's a constant, no?

src/components/IfAllowed.js
```diff
@@ -0,0 +1,65 @@
...
+import PropTypes from 'prop-types';
+
+import Configurable from './Configurable';
+
+type IfAllowedProps = {
+  /**
+   * The role representing a top-level admin, allowed to view any available widget.
+   * Optional, defaulting to "AIE_Administrator"
+   */
+  adminRole: string;
```
lvaldezattivio: I think we can just hard-code this.

src/components/IfAllowed.js
```diff
@@ -0,0 +1,65 @@
...
+        <div>
+          {this.props.children}
+        </div>
+      );
+    }
+    return null;
+  }
+}
+
+export default Configurable(IfAllowed);
```
lvaldezattivio: If we remove the admin role prop, then we don't have to make this a Configurable...

src/index.js
```diff
@@ -69,6 +69,7 @@ export FacetSearchBar from './components/FacetSearchBar';
 export FormattedDate from './components/FormattedDate';
 export GridLayout from './components/GridLayout';
 export Header360 from './components/Header360';
+export IfAllowed from './components/IfAllowed';
```
lvaldezattivio: You'll need to add it to the style guide config and add an example file for it there, too, to get it to build. Also, it's a good way to validate it without having to actually use it in an application. You can probably look for one of the other example markdown files that uses the AuthUtils class to set stuff up since you'll need to fake a logged-in user to do this.

src/components/IfAllowed.js
benmvachon: Okay,
I'll make it a static field in AuthUtils

src/components/IfAllowed.js
benmvachon: true

src/index.js
benmvachon: Okay, I'll look into that

benmvachon:
Can we meet at some point to talk about these scripts/steps for SUIT's build process?
I'm definitely not taking full advantage of them.

src/components/IfAllowed.js
```diff
@@ -0,0 +1,52 @@
...
+
+  static contextTypes = {
+    auth: PropTypes.any,
+  };
+
+  isAllowed(): boolean {
+    let allowed: boolean = false;
+    if (this.context.auth && this.context.auth.state && this.context.auth.state.user) {
+      allowed = (
+        this.context.auth.state.user.roles.indexOf(this.props.requiredRole) >= 0 ||
```
lvaldezattivio: JavaScript's Array class includes an `includes()` method which is nicer than using `indexOf()`.

src/components/AuthRoute.js
```diff
@@ -41,13 +41,17 @@ type AuthRouteState = {
...
-    required: null,
+    requiredRole: null,
     location: null, // This should be filled in by the router
     authType: 'NONE',
   };
 
   static displayName = 'AuthRoute';
 
+  static childContextTypes = {
+    auth: PropTypes.any,
```
lvaldezattivio: Do you want to give children access to the AuthRoute component? Or just the user info object? It might make more sense to include just the user info in the context...

However, in case you really do need the whole AuthRoute component, BTW, if you look at ListSerarchResult, you'll see an example of a better PropTypes usage here... You can do:
```auth: PropTypes.instanceOf(AuthRoute);```
I should really update the rest of the occurrences in the code to do the same since the use of `any` is discouraged and in newer versions of Flow will be a warning or error...
But I digress....

benmvachon: Just the user info,
That makes sense, I'll update it.

PR #73, opened 2018-12-11
Feature/plat 42796

anuroopa18:


src/components/MasterDetails.js
```diff
@@ -82,6 +82,10 @@ type MasterDetailsProps = {
    * Optional; defaults to 0 pixels.
    */
   padding: number;
+  /**
+   * The boolean flag if set, the selection of the row will be retained
```
benmvachon: @luigivaldez what's the right wording for this comment? "A flag for disabling deselection if only one row is selected"?

luigivaldez: I would say something like
```
/*
 * If set, then the component will not allow the user to deselect all rows in the child table.
 * For single-select components, the user will only be able to change the selection to
 * another row; for multi-select components, the user will not be able to deselect the last-
 * selected row without first selecting another one. This means that, unless the rows
 * property is empty, there will always be an item selected in the table and, therefore,
 * shown in the details view.
 */
```
Although, now that I'm thinking about it, as long as the details view (the inspector) can handle the case where nothing is selected (which it has to be able to do anyway, in case there are no connectors in the table), maybe this whole idea of disallowing an empty selection is kind of awkward—it would mean, for multi-select tables, for example, that the user couldn't deselect Connector A and then select Connector B; instead needing to select Connector B first and _then_ deselect Connector A... which seems unnecessarily restrictive. I'm not saying there aren't cases where we want this (especially for single selections) but maybe we shouldn't turn it on here? Hmm…

src/components/Table.js
```diff
@@ -90,6 +90,10 @@ type TableProps = {
    * be re-rendered with new values for the sortColumn and rows properties.
    */
   onSort: null | (sortColumn: number) => void;
+  /**
+   * The boolean flag if set, the selection of the row will be retained
```
luigivaldez: We might want to put the full descriptive comment here and have the comment on the property in the MasterDetails component say something like "See the noEmptySelection property of the Table component for details."

src/components/Table.js
```diff
@@ -178,7 +184,7 @@ export default class Table extends React.Component<TableDefaultProps, TableProps
         selectedRowIds = [rowId];
       }
       this.props.onSelect(selectedRowIds, rowId);
-    } else {
+    } else if (!this.props.noEmptySelection) {
```
luigivaldez: We only want to disable deselection if the current selection has only one item in it, right? The way it is now, it looks like you won't be able to ever deselect anything in a multi-select table...

src/components/Table.js
```diff
@@ -212,6 +218,7 @@ export default class Table extends React.Component<TableDefaultProps, TableProps
     this.props.onSort(colNum);
   }
 
+
```
luigivaldez: (Did you add an extra line here? I think it'll be a lint error...)

src/components/MasterDetails.js
benmvachon: If there are no connectors in the system, or if no connectors match the filter, the `MasterDetails` component isn't displayed at all. Instead there's a message prompting the user to clear the filters or create a connector.

luigivaldez: OK

benmvachon: In this multi-select table, it's not possible to switch the selected connector? I guess I thought that would be the default behavior and that the user would need to hold shift or something in order to select multiple connectors...
Either way (deselecting Connector A and then selecting Connector B or selecting Connector B then deselecting Connector A) it does seem a little restrictive.

benmvachon: If having no selection should be normal, then I would prefer it be handled in the MasterDetails component by default, and that the table would just take up the screen real-estate that the inspector would have.

luigivaldez: The way the out-of-the-box bootstrap table works is that the rows behave like checkboxes when in multi-select mode (the setting is actually called "checkbox"), so just clicking a row will toggle its selected state. I haven't investigated what we would need to do to change that behavior... we don't get information about the event that triggered the selection change, so we don't know if any modifier keys were pressed at the time, but we could probably figure out a way around that and implement a "real" selection algorithm like most list-type controls have (e.g., shift-click selects all items between the current one and the previous one, control-click (command-click) adds/removes the clicked row from the selection, clicking a row without any modifier keys changes the selection to _only_ that one, etc.). That's what I think I'd like to do in the long run... (At that point, I don't know that the react-bootstrap-table library will be buying us much... so maybe we would just write our own component altogether. 😕)

As for the other comment, I don't think we want the table to grow/shrink based on whether there's a selection, as that would be very distracting for it to constantly grow and shrink as the selection changes... In addition, I can imagine that some details components might want to say something about the empty selection (maybe a summary or a message saying "Hey, dude, select something!").

src/components/MasterDetails.js
```diff
@@ -167,6 +169,9 @@ export default class MasterDetails extends React.Component<MasterDetailsDefaultP
     const tableWidth = this.props.split;
     const detailsWidth = 12 - tableWidth;
     const halfPadding = `${this.props.padding / 2}px`;
+    const detailsRow = this.props.rows.find((row) => {
+      return row.id === this.state.detailsRow.id;
```
benmvachon: `detailsRow` should be removed from the state and replaced with a string called `detailsRowId` since the Id is the only part of it that we ever use

PR #76, opened 2018-12-18
Feature/plat 42796

anuroopa18:


src/components/MasterDetails.js
```diff
@@ -137,6 +137,31 @@ export default class MasterDetails extends React.Component<MasterDetailsDefaultP
   }
   state: MasterDetailsState;
 
+  componentWillReceiveProps(newProps: MasterDetailsProps) {
```
luigivaldez: [I should have remembered this earlier, but componentWillReceiveProps() is deprecated and will go away in React 17. I wouldn't change anything now... It's still okay to use it, but we'll eventually need to update this to be a different lifecycle method, like, for example, componentDidUpdate...]

src/components/MasterDetails.js
```diff
@@ -137,6 +137,31 @@ export default class MasterDetails extends React.Component<MasterDetailsDefaultP
...
+    const isRowPresent = newProps.rows.find((row) => {
+      return (row.id === this.state.detailsRowId);
+    });
```
luigivaldez: We actually need to check all of the IDs in this.state.selectedRows to see if they're in the no longer in the list of rows, and remove them if they're not, regardless of multi-select.

Once the selectedRows array is up to date, we can look at the multiSelect flag: if multiSelect is true and selectedRows has no elements, we would want to add the ID of the first item from props.rows (if props.rows is an empty list, then there's nothing we can do to ensure there's a selection—in this case selectedRows would still be empty.

Once we've updated selectedRows, we can figure out which ID to use for the detailsRowId. If selectedRows is empty, then detailsRowId will be the empty string. If it's not empty, detailsRowId will be the last item in selectedRows. (Note that if the previous row being shown in the details component is still in the list of rows, then it will still end up being detailsRowId and the details view won't change unless the row's data has changed.)

Finally, after all of that, we can update the state with the new list of selectedRows and the new detailsRowId.

src/components/MasterDetails.js
```diff
@@ -137,6 +137,31 @@ export default class MasterDetails extends React.Component<MasterDetailsDefaultP
...
+    if (!isRowPresent) {
+      if (this.props.multiSelect) {
+        const oldPosition = this.state.selectedRows.indexOf(this.state.detailsRowId);
+        if (oldPosition >= 0) {
+          // If it was in the previous array, remove it...
+          this.state.selectedRows.splice(oldPosition, 1);
```
luigivaldez: Don't forget that we cannot modify the state like this... We need to call this.setState() with the new state... We should make a copy of the old selectedRows and modify it and then include it in the setState call (together with the updated detailsRowId).

src/components/MasterDetails.js
```diff
@@ -137,6 +137,31 @@ export default class MasterDetails extends React.Component<MasterDetailsDefaultP
   }
   state: MasterDetailsState;
 
+  componentWillReceiveProps(newProps: MasterDetailsProps) {
+    const isRowPresent = newProps.rows.find((row) => {
```
luigivaldez: (As a short cut, I think you can see if the props.rows array is empty and then just set selectedRows to [] and detailsRowId to '' and be done with it...)

src/components/MasterDetails.js
```diff
@@ -138,25 +138,34 @@ export default class MasterDetails extends React.Component<MasterDetailsDefaultP
...
       if (this.props.multiSelect) {
-        const oldPosition = this.state.selectedRows.indexOf(this.state.detailsRowId);
-        if (oldPosition >= 0) {
-          // If it was in the previous array, remove it...
-          this.state.selectedRows.splice(oldPosition, 1);
-          const detailsRowId = this.state.selectedRows.length > 0 ? this.state.selectedRows[this.state.selectedRows.length - 1] : '';
-          this.setState({
-            detailsRowId,
-          });
+        const updatedSelectedRows = newSelectedRows.length > 0 ? newSelectedRows : newProps.rows[0].id;
```
benmvachon: This should only happen if `noEmptySelection` is `true`
Otherwise, it's okay for `updatedSelectedRows` to be an empty array.

src/components/MasterDetails.js
```diff
@@ -138,25 +138,34 @@ export default class MasterDetails extends React.Component<MasterDetailsDefaultP
...
+        this.setState({
+          selectedRows: updatedSelectedRows,
+        });
+        let newDetailsRowId;
+        if (updatedSelectedRows.includes(this.state.detailsRowId)) {
+          newDetailsRowId = this.state.detailsRowId;
+        } else {
+          newDetailsRowId = updatedSelectedRows[newSelectedRows.length - 1];
```
benmvachon: This should also only happen if `noEmptySelection` is `true`
Otherwise, it's okay for `newDetailsRowId` to be either `null` or an empty string. I'm not sure which...

src/components/MasterDetails.js
```diff
@@ -138,25 +138,34 @@ export default class MasterDetails extends React.Component<MasterDetailsDefaultP
...
+    if (newProps.rows.length === 0) {
+      this.setState({
+        selectedRows: [],
+        detailsRowId: '',
+      });
+    } else {
+      const ids = newProps.rows.map((row) => {
+        return row.id;
+      });
+      const newSelectedRows = this.state.selectedRows.filter((row) => { return ids.includes(row); });
```
luigivaldez: Nice!

src/components/MasterDetails.js
```diff
@@ -138,25 +138,34 @@ export default class MasterDetails extends React.Component<MasterDetailsDefaultP
...
       if (this.props.multiSelect) {
```
luigivaldez: Yes, what Ben said is correct... I think the `if` statement could be changed to `if (this.props.noEmptySelection)` instead because if it's set, you don't care about multi-select, you always want to make a selection.
Actually, you can change it to `if (this.props.noEmptySelection && newSelectedRows.length === 0)` and then you can push the first row's ID onto the newSelectedRows array. After that, outside of the `if` statement's block, you can determine the detailsRowId (it's the last item in the newSelectedRows array, unless the array is empty, in which case it's the empty string) and then update the state with these two values (at the same time, rather than calling `setState()` twice).

src/components/MasterDetails.js
luigivaldez: (@benmvachon, we've been using '' instead of null for the `detailsRowId`.)

src/components/MasterDetails.js
```diff
@@ -148,26 +148,17 @@ export default class MasterDetails extends React.Component<MasterDetailsDefaultP
...
-        let newDetailsRowId;
-        if (updatedSelectedRows.includes(this.state.detailsRowId)) {
-          newDetailsRowId = this.state.detailsRowId;
-        } else {
-          newDetailsRowId = updatedSelectedRows[newSelectedRows.length - 1];
-        }
-        this.setState({
-          detailsRowId: updatedSelectedRows.length > 0 ? newDetailsRowId : '',
-        });
+      let updatedSelectedRows = [];
```
luigivaldez: Any reason why you have a new variable here instead of just pushing `newProps.rows[0].id ` onto `newSelectedRows`?

src/components/MasterDetails.js
```diff
@@ -148,26 +148,17 @@ export default class MasterDetails extends React.Component<MasterDetailsDefaultP
...
+      if (this.props.noEmptySelection && newSelectedRows.length === 0) {
+        updatedSelectedRows.push(newProps.rows[0].id);
       } else {
-        this.setState({
-          selectedRows: newSelectedRows.length > 0 ? newSelectedRows : newProps.rows[0].id,
-          detailsRowId: newSelectedRows.length > 0 ? newSelectedRows[0] : newProps.rows[0].id,
-        });
+        updatedSelectedRows = newSelectedRows;
       }
+      const newDetailsRowId = (updatedSelectedRows.length > 0) ? (updatedSelectedRows[newSelectedRows.length - 1]) : '';
```
luigivaldez: I think you want to use the same variable in the second half of this, right? Either `updatedSelectedRows[updatedSelectedRows.length - 1]` or `newSelectedRows[newSelectedRows.length - 1]` (well, the latter if you eliminate `updatedSelectedRows` entirely).

src/components/MasterDetails.js
anuroopa18: Done

src/components/MasterDetails.js
anuroopa18: Done.

PR #79, opened 2018-12-20
PLAT-42707 New field added in the CardPicker and used same in CardPic…

chinmay-jomraj:
…kerItem

### DIFF
--- a/src/components/CardPickerCard.js
+++ b/src/components/CardPickerCard.js
@@ -18,6 +18,10 @@ type CardPickerCardProps = {
    * The icon is limited in size to 60x60 pixels.
    */
   defaultIconUri: string | null;
+   /**
+   * The description for scanner type in the card.
+   */
+  desc: string | null;
   /**
    * If set, then the card is drawn with a selected border and is not
    * clickable.
@@ -100,7 +104,7 @@ export default class CardPickerCard extends React.Component<CardPickerCardDefaul
     );
 
     return (
-      <div
+      <div title={this.props.desc}
         style={style}
         onClick={this.onClick}
         role="button"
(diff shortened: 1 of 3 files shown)


src/components/CardPickerCard.js
```diff
@@ -18,6 +18,10 @@ type CardPickerCardProps = {
    * The icon is limited in size to 60x60 pixels.
    */
   defaultIconUri: string | null;
+   /**
+   * The description for scanner type in the card.
+   */
+  desc: string | null;
```
luigivaldez: Can we rename "desc" to "description"? Both here in the component and in the CardPickerItem class too.

In addition, the comment shouldn't mention scanner types because the SUIT component is a generic card that can be used for anything.

luigivaldez: We also need to add the description property to the CardPickerCardDefaultProps type (around line 45), and then add it to the defaultProps variable (around line 52) in the class (setting it to null) so the value will be clear if the parent component doesn't set it.

chinmay-jomraj: Can we put following comment for the description field ?
-A descriptive summary to supplement the card contents

luigivaldez: I think we're going to eventually make the description visible at all times instead of being a tooltip... maybe just say "An optional description of the card's contents"?

src/components/CardPickerCard.js
```diff
@@ -19,9 +19,9 @@ type CardPickerCardProps = {
    */
   defaultIconUri: string | null;
    /**
-   * The description for scanner type in the card.
+   * The descriptive summary to suppliment the card contents.
```
benmvachon: tiny thing: "suppliment" is a typo, it should say "supplement"

PR #83, opened 2019-01-10
PLAT-42796 fix MasterDetails/Table 'noEmptySelection' flag handling

benmvachon:


benmvachon:
This is all to fix the issue in ConnectorAdmin where empty row selection was still being allowed.
I moved any `noEmptySelection` handling in `MasterDetails` down into `Table`.
I removed the `isSelected` parameter from `Table.rowSelect` because it was redundant information.
I figured out that what `BootstrapTable` does _after_ its `onSelect` handler depends on an optional `boolean` return value and made `Table.rowSelect` return an appropriate `boolean` value.

benmvachon:
I am going to add the style related changes in to this PR as well, by the way. So no rush on the review.

src/components/Table.js
```diff
@@ -172,38 +172,39 @@ export default class Table extends React.Component<TableDefaultProps, TableProps
     }
   }
 
-  rowSelect(rowData: any, isSelected: boolean) {
-    const rowId = rowData.id;
+  rowSelect(rowData: any): boolean {
```
luigivaldez: So, when a row is clicked and the table is a multi-select one, we're toggling the selection state of that row, unless its currently selected and the `noEmptySelection` flag is set, in which case we don't change anything...

In the case of a single-select table, however, if the `noEmptySelection` is not set, it seems like the user can not deselect the selected row. Is that the case? And is that intentional?

luigivaldez: Also, I realize it's not part of this merge _per se_, but the comment for the `noEmptySelection` property is written as if it were the `noEmptySelection` property of the MasterDetails component instead of the Table component...
(Also, if we're saying that you can't ever deselect a row in a single-select table once one has been selected, then the `noEmptySelection` flag really only applies to multi-select tables and, for all intents and purposes, it's always true for single-select ones, and we should say that in the comment if we decide that this is how we want it to work...

benmvachon: That wasn't my intention.
I'll fix it so that `noEmptySelection` is necessary to have this behavior in single-select as well.

src/components/MasterDetails.js
```diff
@@ -56,6 +56,11 @@ type MasterDetailsProps = {
    * in the Table component for more details.
    */
   onSort: null | (sortColumn: number) => void;
+  /**
```
luigivaldez: Cool... this is more convenient than having to do the wrapped component on the outside... Can you move it so it's right under the details property? Also, maybe tweak the comment to say something like "Additional properties to pass to the details component in addition to the `data` property. Optional."

src/components/MasterDetails.js
```diff
@@ -207,12 +222,14 @@ export default class MasterDetails extends React.Component<MasterDetailsDefaultP
...
                 noEmptySelection={this.props.noEmptySelection}
+                selectedClassName={this.props.selectedClassName}
+                tableClassName={this.props.tableClassName}
               />
               {this.props.footer}
             </div>
           </Col>
           <Col lg={detailsWidth} md={detailsWidth} sm={12} xs={12} style={{ paddingLeft: halfPadding, paddingRight: 0 }}>
-            <Detail data={detailsRow} />
+            <Detail data={detailsRow} {...this.props.detailsProps} />
```
benmvachon: @luigivaldez ,
one more sort of unexpected change...
The `Details` component in the connector admin ui needs more properties than just `data` in order to function.
My understanding is that the original intention would be to pass `details` as a function taking props as a parameter.
e.g.
```jsx
          <MasterDetails
            columns={this.getTableColumns()}
            rows={rows}
            details={(props) => {
              return (
                <InspectorDetails
                  connector={props.data}
                  scannerTypes={scannerTypes}
                  isAdmin={isAdmin}
                  editConnector={editConnector}
                  openPrompt={openPrompt}
                  preview={preview}
                />
              );
            }}
```

This did work, but the problem was that the `InspectorDetails` component didn't seem to be properly maintained in the virtual DOM. Whenever a render happened, the `Details` component wouldn't just re-render, it's whole constructor would be called again which led to deeper functional issues.
This `detailsProps` property was my idea to fix the issue. If you think there is a better way, or if I'm doing something wrong, just let me know.

src/components/MasterDetails.js
```diff
@@ -222,7 +229,7 @@ export default class MasterDetails extends React.Component<MasterDetailsDefaultP
...
```
luigivaldez: Note that if the detailsProps contains a "data" member, it will override the one you're explicitly passing here. (In JSX, the spread works the same as in JavaScript itself where items to the right override things that precede them.) I think you can just change the order, or you can use the Object.assign() method and then spread the result...

benmvachon:
@luigivaldez I wanted to try the npm publishing stuff, so

1. I merged this in
2. checked out and pulled `master`
3. ran `npm run updateversion`
4. got an error saying I can't push to `master`
5. checked out my branch again
6. ran `npm run updateversion` (successfully)
7. ran `npm run publishlib`
8. got an error saying I'm not authorized

I kind of expected it not to work, but I wanted to try it out anyway.

PR #87, opened 2019-02-07
Bug/plat 42805

klk1010:
Addresses [PLAT-42805](https://jira.attivio.com/browse/PLAT-42805).

Adds an additional optional prop `homeUrl` to `<Masthead />`.
If the prop is specified, the logo becomes an `<a />` instead of a `<button />`.
If not specified, the `homeRoute` is used.
If neither are specified, the default `/` for `homeRoute` is used.
If both are specified `homeUrl` takes precedence.

Modified `componentWillReceiveProps` to only update the user when the username prop changes. This reduces unnecessary renders by about 10x.

Changed `componentWillMount` to `componentDidMount`. `componentWillMount` is deprecated and should never have been used to fetch data from the server or set state. `componentDidMount` will serve the same purpose in a safer way. This could use more attention, but is beyond the scope of this PR.
See https://reactjs.org/blog/2018/03/27/update-on-async-rendering.html for more information on recommended practices for async rendering.

### DIFF
--- a/src/components/Masthead.js
+++ b/src/components/Masthead.js
@@ -129,7 +129,7 @@ class Masthead extends React.Component<MastheadDefaultProps, MastheadProps, Mast
     }
   }
 
-  navigateHome() {
+  navigateHome = () => {
     if (this.homeLink) {
       this.homeLink.blur();
     }
--- a/src/components/Masthead.js
+++ b/src/components/Masthead.js
@@ -18,7 +18,8 @@ type MastheadProps = {
   logoAlt: string | null;
   /** The route to navigate to when the user clicks the logo. Defaults to '/' */
   homeRoute: string | null;
-  /** The optional url to navigate to when the user clicks the logo. If homeRoute
+  /**
+   * The optional url to navigate to when the user clicks the logo. If homeRoute
    *  and homeUrl are both specified, homeUrl takes precedence.
    */
   homeUrl: string | null;
@@ -52,7 +53,7 @@ type MastheadDefaultProps = {
   logoUri: string | null;
   logoAlt: string | null;
   homeRoute: string | null;
-  homeUrl: ?string;
+  homeUrl: string | null;
   applicationName: string | null;
   multiline: boolean;
   searchEngineType: 'attivio' | 'solr' | 'elastic';
--- a/src/components/Masthead.js
+++ b/src/components/Masthead.js
@@ -48,6 +48,7 @@ type MastheadDefaultProps = {
   logoUri: string | null;
   logoAlt: string | null;
   homeRoute: string | null;
+  homeUrl: ?string;
   applicationName: string | null;
   multiline: boolean;
   searchEngineType: 'attivio' | 'solr' | 'elastic';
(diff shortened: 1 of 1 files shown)


src/components/Masthead.js
```diff
@@ -18,6 +18,10 @@ type MastheadProps = {
   logoAlt: string | null;
   /** The route to navigate to when the user clicks the logo. Defaults to '/' */
   homeRoute: string | null;
+  /** The optional url to navigate to when the user clicks the logo. If homeRoute
+   *  and homeUrl are both specified, homeUrl takes precedence.
```
benmvachon: Typically, when one of these comments spans multiple lines, we start it on the second line, leaving the first and last lines blank.
So the comment would look like
```
  /**
   * The optional url to navigate to when the user clicks the logo. If homeRoute
   * and homeUrl are both specified, homeUrl takes precedence.
   */
```
instead of
```
  /** The optional url to navigate to when the user clicks the logo. If homeRoute
   *  and homeUrl are both specified, homeUrl takes precedence.
   */
```

klk1010: Good to know, I'll update that. Once we update to the latest eslint version we can add this rule to our eslint config to enforce this:
https://eslint.org/docs/rules/multiline-comment-style#enforce-a-particular-style-for-multiline-comments-multiline-comment-style

luigivaldez: This convention is really only for JSDoc-style comments (and comes from similar conventions for JavaDoc comments). Apparently the ESLint folks are deprecating the current rules they have for JSDoc (see https://eslint.org/blog/2018/11/jsdoc-end-of-life) in favor of an ESLint plug-in to handle it (https://github.com/gajus/eslint-plugin-jsdoc). However, I don't see an option there for this sort of rule (e.g., enforcing that multi-line JSDoc comments have a newline right after the first /**... although there may be others we might want (e.g. newline-after-description or check-tag-names).

In any case, I don't think we want to use the multiline-comment-style rule because it doesn't differentiate between JSDoc comments and just plain block-style comments. It would, for example, complain about two consecutive lines with // comments... which, I think should be fine.

E.g.:
```
// this is a long comment
// separated into two lines
```

or
```
// This is one comment
// And this is another one that just happens to be adjacent to the first one
```

Anyway... :)

src/components/Masthead.js
```diff
@@ -48,6 +52,7 @@ type MastheadDefaultProps = {
   logoUri: string | null;
   logoAlt: string | null;
   homeRoute: string | null;
+  homeUrl: ?string;
```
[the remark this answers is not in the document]
klk1010: oops, I'll make them match.

One thought - the `string | null` typing is a bit odd. I won't break away from that here as a one off, but if a prop is not required, I would not expect to define it and pass null, I would just omit it, `null` and `undefined` being different. Since proptypes and flow define required in different ways, I would expect to see in proptypes: `myProp: string;` and then in the flow type: `myProp: ?string` to use the [maybe type](https://flow.org/en/docs/types/primitives/#toc-maybe-types). I can always pass an empty string if need be, like if I'm defining a prop, but maybe it's dependent on other data being available.

(What I did is still mismatched and needs to be corrected.)

src/components/Masthead.js
klk1010: Thanks for the clarification!

PR #88, opened 2019-02-22
Feat/plat 42929

klk1010:
Addresses [PLAT-42929](https://jira.attivio.com/browse/PLAT-42929).

Adds support for keyboard multiselect in Table and MasterDetails using shift key for selecting multiple rows and ctrl/meta key for selecting individual rows. 
This new selection support uses the concept of an anchor row. There can only be one anchor row. It may or may not be the same as the active row displayed in MasterDetails. The anchor row will always be the last row clicked without a modifier or with the ctrl key. Clicks made with the shift key do not change the anchor row. The anchor row is used to determine what rows will be selected when the shift key is used for multiselect.
The selections behavior largely follows Outlooks selection behavior with one intentional difference.
See the following pdfs for detailed differences.
[Outlook Selection Behavior.pdf](https://github.com/attivio/suit/files/2911116/Outlook.Selection.Behavior.pdf)
[Table Selection Behavior.pdf](https://github.com/attivio/suit/files/2911117/Table.Selection.Behavior.pdf)

This is a breaking change for the Table multiselect behavior. Clicking a row will no longer make multiple selections unless a ctrl/meta or shift key is pressed down.

If multiselect is true, there as an additional optional prop that may be passed to specify the background color for the row corresponding to the current detail row being displayed. This change is backwards compatible and opt-in. If users omit the new prop, the current behavior won't change.

I updated the table header prop to support either a render function or a component as it does today. Users passing in a component can continue to do so.

Performance improvements: 
Updates the Configurable component to no longer use `componentWillReceiveProps`. Because configuration options do not change, setting these on component mount is sufficient. The use of that lifecycle function was triggering cascading updates, which frequently resulted in an infinite cycle of rerenders. Anytime any child component received an update (triggered by state or prop changes), componentWillReceiveProps was called and unconditionally rewrote the incoming props to state, triggering a subsequent state update, and a subsequent call of componentWillReceiveProps.
Because of the wide use of this component, this should be a significant performance improvement. However, most use cases won't notice a change. Only areas using virtualization or managing large data sets, or data hungry loads such as video would likely see a difference.
This change is backwards compatible and is invisible to library consumers - configuration can still be specified as it is today.

I updated references to `componentWillMount` (which is being deprecated) to instead use `componentDidMount` and added `FIXME` notations to various usages of `componentDidMount` where state is being set in that lifecycle method and to instances where `componentWillReceiveProps` is setting state without first doing a prop change comparison.
https://reactjs.org/blog/2018/03/27/update-on-async-rendering.html

### DIFF
--- a/src/components/Configurable.js
+++ b/src/components/Configurable.js
@@ -28,12 +28,9 @@ function Configurable(WrappedComponent: React.Component): React.Component {
       };
     }
 
+    // FIXME: Never set state in componentDidMount
     componentDidMount() {
       this.fillWithDefaults(this.props);
-    }
-
-    componentWillReceiveProps(nextProps: any) {
-      this.fillWithDefaults(nextProps);
     }
 
     fillWithDefaults(props: any) {
@@ -46,6 +43,7 @@ function Configurable(WrappedComponent: React.Component): React.Component {
           filled[property] = this.context.configuration.get(name, property, props[property]);
         });
       }
+      console.log('<Configurable /> fillWithDefaults() : ', WrappedComponent.displayName);
       this.setState({
         filled: true,
         props: filled,
--- a/src/components/Configurable.js
+++ b/src/components/Configurable.js
@@ -1,5 +1,6 @@
 import React from 'react';
 import PropTypes from 'prop-types';
+import ObjectUtils from '../util/ObjectUtils';
 
 type ConfigurableState = {
   filled: boolean;
@@ -31,6 +32,12 @@ function Configurable(WrappedComponent: React.Component): React.Component {
     // FIXME: Never set state in componentDidMount
     componentDidMount() {
       this.fillWithDefaults(this.props);
+    }
+
+    componentWillReceiveProps(nextProps: any) {
+      if (!ObjectUtils.deepEquals(this.props, nextProps)) {
+        this.fillWithDefaults(nextProps);
+      }
     }
 
     fillWithDefaults(props: any) {
--- a/src/components/Configurable.js
+++ b/src/components/Configurable.js
@@ -43,7 +43,7 @@ function Configurable(WrappedComponent: React.Component): React.Component {
           filled[property] = this.context.configuration.get(name, property, props[property]);
         });
       }
-      console.log('<Configurable /> fillWithDefaults() : ', WrappedComponent.displayName);
+
       this.setState({
         filled: true,
         props: filled,
--- a/src/components/ConfirmationDialog.js
+++ b/src/components/ConfirmationDialog.js
@@ -74,8 +74,8 @@ export default class ConfirmationDialog extends React.Component<ConfirmationDial
 
   render() {
     const okButtonStyle = this.props.dangerous ? 'danger' : 'primary';
-    return this.props.show ? (
-      <Modal show>
+    return (
+      <Modal show={this.props.show} onHide={this.props.onCancel}>
         <Modal.Header>
           <Modal.Title>{this.props.title}</Modal.Title>
         </Modal.Header>
@@ -87,6 +87,6 @@ export default class ConfirmationDialog extends React.Component<ConfirmationDial
           <Button onClick={this.props.onConfirm} bsStyle={okButtonStyle}>{this.props.confirmButtonLabel}</Button>
         </Modal.Footer>
       </Modal>
-    ) : null;
+    );
   }
 }
--- a/src/components/MasterDetails.js
+++ b/src/components/MasterDetails.js
@@ -1,6 +1,6 @@
 // @flow
 
-import React from 'react';
+import * as React from 'react';
 
 import Grid from 'react-bootstrap/lib/Grid';
 import Row from 'react-bootstrap/lib/Row';
@@ -68,13 +68,13 @@ type MasterDetailsProps = {
    * user filter the list of items in the table. May also be a function that uses selectedRows and
    * renders a header component. Optional; if not set, the table will be flush with the top of the details component.
    */
-  header: any;
+  header?: null | React.Node | (selectedRows: Array<string>) => React.Node;
   /**
    * A footer component that can be inserted below the table (but to the left of the
    * details component). For example, this might be used to include controls to paginate
    * the table. Optional.
    */
-  footer: any;
+  footer?: null | React.Node | (selectedRows: Array<string>) => React.Node;
   /**
    * This determines the ratio of table to details, based on Bootstrap's 12-column grid.
    * This is the width of the table; the details component will use the remainder. Optional;
@@ -104,33 +104,34 @@ type MasterDetailsProps = {
    * The class name to apply to the table element. Optional.
    */
   tableClassName: string;
-  /** Optional background color to apply to the last selected row. Only used if multiSelect is specified as well. Takes precedence
-   *  over all other background colors specified through classNames.
+  /**
+   * Optional background color to apply to the last selected row. Only used if multiSelect is specified as well.
+   * Takes precedence over all other background colors specified through classNames.
    */
   anchorRowBackgroundColor: ?string;
 };
 
 type MasterDetailsDefaultProps = {
   detailsProps: any;
+  footer?: null | React.Node | (selectedRows: Array<string>) => React.Node;
+  header?: null | React.Node | (selectedRows: Array<string>) => React.Node;
   multiSelect: boolean;
-  onSelect: null | (rowsIds: Array<string>, newlySelectedRowId: string | null) => void;
-  sortColumn: number;
-  onSort: null | (sortColumn: number) => void;
-  header: any;
-  footer: any;
-  split: ColumnCount;
-  padding: number;
   noEmptySelection: boolean;
-  tableContainerClassName: string | void;
+  onSelect: null | (rowsIds: Array<string>, newlySelectedRowId: string | null) => void;
+  onSort: null | (sortColumn: number) => void;
+  padding: number;
   selectedClassName: string;
+  sortColumn: number;
+  split: ColumnCount;
   tableClassName: string;
+  tableContainerClassName: string | void;
 };
 
 type MasterDetailsState = {
-  /** The IDs of the selected rows */
-  selectedRows: Array<string>;
   /** The id for the most-recently selected row, to be displayed in the details pane. */
   detailsRowId: string;
+  /** The IDs of the selected rows */
+  selectedRows: Array<string>;
 };
 
 /**
@@ -141,18 +142,18 @@ type MasterDetailsState = {
 export default class MasterDetails extends React.Component<MasterDetailsDefaultProps, MasterDetailsProps, MasterDetailsState> {
   static defaultProps = {
     detailsProps: {},
-    multiSelect: false,
-    onSelect: null,
-    sortColumn: 0,
-    onSort: null,
-    header: null,
     footer: null,
-    split: 8,
-    padding: 0,
+    header: null,
+    multiSelect: false,
     noEmptySelection: false,
-    tableContainerClassName: undefined,
+    onSelect: null,
+    onSort: null,
+    padding: 0,
     selectedClassName: 'attivio-table-row-selected',
+    sortColumn: 0,
+    split: 8,
     tableClassName: 'table table-striped attivio-table attivio-table-sm',
+    tableContainerClassName: undefined,
   };
 
   static displayName = 'MasterDetails';
@@ -174,7 +175,7 @@ export default class MasterDetails extends React.Component<MasterDetailsDefaultP
       if (newProps.rows.length === 0) {
         this.setState({
           selectedRows: [],
-          detailsRowId: null,
+          detailsRowId: '',
         });
       } else {
         const ids = newProps.rows.map((row) => {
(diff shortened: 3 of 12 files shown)


klk1010:
Updating Table multiselect behavior to only select if the ctrl key is held down. Closing until ready for review.

src/components/MasterDetails.js
```diff
@@ -103,6 +104,10 @@ type MasterDetailsProps = {
    * The class name to apply to the table element. Optional.
    */
   tableClassName: string;
+  /** Optional background color to apply to the last selected row. Only used if multiSelect is specified as well. Takes precedence
+   *  over all other background colors specified through classNames.
+   */
+  anchorRowBackgroundColor: ?string;
```
luigivaldez: I'm actually wondering we should use the normal select color for the anchor row and use the extra color for the selected-but-not-showing-in-the-inspector rows... This would mean, for example, that a multi-select table with one row selected would look the same as a single-select t table with one row selected... what do you think?

src/components/MasterDetails.js
```diff
@@ -124,7 +129,7 @@ type MasterDetailsDefaultProps = {
 type MasterDetailsState = {
   /** The IDs of the selected rows */
```
luigivaldez: It might be useful to add a comment here about selectedRows having the rows being in chronological order with the most recently selected one at the end of the list... (if this is still a true statement, of course).

src/components/MasterDetails.js
```diff
@@ -155,32 +160,37 @@ export default class MasterDetails extends React.Component<MasterDetailsDefaultP
...
-      const newSelectedRows = this.state.selectedRows.filter((row) => { return ids.includes(row); });
-      if (this.props.noEmptySelection && newSelectedRows.length === 0) {
-        newSelectedRows.push(newProps.rows[0].id);
+    const rowsChanged = !ObjectUtils.arrayEquals(newProps.rows, this.props.rows);
+
+    if (rowsChanged) {
+      if (newProps.rows.length === 0) {
+        this.setState({
+          selectedRows: [],
+          detailsRowId: null,
```
luigivaldez: Does Flow not complain about setting detailsRowId to null when the type is specified as `string` in MasterDetailsState?

src/components/MasterDetails.js
```diff
@@ -124,7 +129,7 @@ type MasterDetailsDefaultProps = {
 type MasterDetailsState = {
   /** The IDs of the selected rows */
   selectedRows: Array<string>;
-  /** The object for the most-recently selected row, to be displayed in the details pane.  */
+  /** The id for the most-recently selected row, to be displayed in the details pane. */
   detailsRowId: string;
```
luigivaldez: If the selectedRows array is always in chronological order, is the `detailsRowId` just here as a convenience? Would be be better to have a method to call to get the details row ID so we don't have to make sure we always keep these two fields in sync.

src/components/Table.js
```diff
@@ -177,81 +197,231 @@ export default class Table extends React.Component<TableDefaultProps, TableProps
...
+      document.addEventListener('keyup', this.keyUp);
+    }
+  }
+
   componentWillReceiveProps(newProps: TableProps) {
     if (!ObjectUtils.arrayEquals(newProps.rows, this.props.rows)) {
       // Reset the sorted rows if the actual rows have changed.
       this.setState({
-        sortedRows: newProps.rows,
+        sortedRows: newProps.rows && newProps.rows.length > 0
```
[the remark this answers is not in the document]
klk1010: Good point, I need to make sure the selection is cleared on search/sort per our discussions on Friday.

src/components/Configurable.js
```diff
@@ -28,14 +28,11 @@ function Configurable(WrappedComponent: React.Component): React.Component {
       };
     }
 
+    // FIXME: Never set state in componentDidMount
     componentDidMount() {
       this.fillWithDefaults(this.props);
     }
 
-    componentWillReceiveProps(nextProps: any) {
```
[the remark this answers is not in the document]
klk1010: Understood, I'll make that change.

klk1010:
It's possible to do it, but it's much more straightforward to track the selection by index. It makes it easier to track group selections with the shift key.

src/components/Configurable.js
```diff
@@ -28,12 +29,15 @@ function Configurable(WrappedComponent: React.Component): React.Component {
...
     }
 
+    // FIXME: Never set state in componentDidMount
     componentDidMount() {
       this.fillWithDefaults(this.props);
     }
 
     componentWillReceiveProps(nextProps: any) {
-      this.fillWithDefaults(nextProps);
+      if (!ObjectUtils.deepEquals(this.props, nextProps)) {
```
klk1010: Still not crazy about this - it's expensive to do a deepEquals comparison every time any state/props change, but I'll leave it be for the time being and come back to it when I can focus on this problem by itself. This is still an improvement as it breaks the loop where this lifecycle would set state unconditionally, infinitely re-triggering itself.

src/components/MasterDetails.js
klk1010: With multiselect, determining the active row is now more complex than chronological order.
A row can only become the active row if it was clicked without a modifier.
A row can only become an anchor row if it was clicked without using the shift key.
Because of this, the details row is tracked separately. Pulling out the oldest selected row won't tell you how that key was clicked. It's currently selected, but it may never have qualified as an active row - not if `noEmptySelection` is `false`. Instead, the `detailsRowId` is updated whenever a qualifying click is made. This is an index driven multiselect. Chunks of rows may be selected at a time. Directionality is determined by the anchor row. Chronological order is no longer the driving factor and is not guaranteed as Sets are not ordered and cannot guarantee order.

src/components/MasterDetails.js
klk1010: I did it this way to reduce the impact of potentially existing tables. Anyone using the table today won't see a style difference. If we're comfortable with forcing that styling change, I'm fine with that.

src/components/MasterDetails.js
klk1010: I'm not getting inline flow feedback, but flow did yell at me with the command line check - fixing it.

src/components/MasterDetails.js
klk1010: https://github.com/attivio/suit/pull/88/files#r262257484

src/components/ConfirmationDialog.js
```diff
@@ -74,8 +74,8 @@ export default class ConfirmationDialog extends React.Component<ConfirmationDial
 
   render() {
     const okButtonStyle = this.props.dangerous ? 'danger' : 'primary';
-    return this.props.show ? (
-      <Modal show>
+    return (
+      <Modal show={this.props.show} onHide={this.props.onCancel}>
```
benmvachon: Why was this `onHide` hook added? The `onCancel` prop is already passed into the `onClick` hook of the cancel button on line 86. Will this cause the `onCancel` callback to execute twice?

src/components/Table.js
```diff
@@ -274,21 +446,41 @@ export default class Table extends React.Component<TableDefaultProps, TableProps
...
-      selectedRowIds = [this.props.selection];
+    if (Array.isArray(selection)) {
+      selectedRowIds = selection;
+    } else if (typeof selection === 'string') {
+      selectedRowIds = [selection];
     } else {
       selectedRowIds = [];
     }
 
+    // FIXME: Move data manipulation outside of render function.
```
benmvachon: Couldn't this be solved by just updating `selectedRowIds` where `selectedRowIndices` is being updated and getting rid of the `selectedRowIndices` set?

src/components/ConfirmationDialog.js
klk1010: I updated this to use the built in show functionality and also eliminated some console warnings I saw. This won't call the `onCancel` function twice. The `onHide` function is only called if you click a non-static backdrop or the "X" button. The Cancel button does not trigger onHide, it's just a button that calls whatever function you pass to it. In this case, that function also happens to be `onCancel`. https://react-bootstrap.github.io/components/modal/#modals-live

src/components/Table.js
```diff
@@ -274,21 +446,41 @@ export default class Table extends React.Component<TableDefaultProps, TableProps
...
+    // If multiSelect include row index selections made using keyboard
```
luigivaldez: Shouldn't all the rows that are selected be in `this.props.selection`? The source of truth for the selection is the property that the parent passes in, right? I thought that the only thing our state keeps is information about the anchor row and the active row, but that those rows should be in the selection array.

src/components/MasterDetails.js
klk1010: Can we also know that no customers are using this? Here, the change is pretty minor and the impact as well so I'll go ahead and change it here, but generally speaking, are we concerned about the impact outside of the organization?

src/components/Table.js
```diff
@@ -177,81 +197,233 @@ export default class Table extends React.Component<TableDefaultProps, TableProps
...
+      onSelect([rowData.id], rowData.id);
+      return true;
+    }
+
+    // multiSelect is not enabled
+    // if this row is already the currently selected row,
+    if (activeRowId === rowData.id) {
       // if we don't allow empty selection
-      if (this.props.noEmptySelection) {
+      if (noEmptySelection) {
```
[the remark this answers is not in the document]
klk1010: It's an odd use case to imagine, either way. If multiselect isn't enabled, being required to use control to deselect is not at all obvious. When multiselect is enabled, it makes sense, but for the single selection case, not so much. Do we have any current use cases for this scenario to think this through on?

src/components/Table.js
```diff
@@ -177,81 +197,233 @@ export default class Table extends React.Component<TableDefaultProps, TableProps
...
+    if (!rowData) {
+      return false;
     }
 
-    // if this row is the currently selected row,
-    if (selectedRowIds === rowId) {
+    if (multiSelect) {
+      if (shiftKeyPressed) {
+        if (anchorRowIndex === null) {
+          this.setState({
```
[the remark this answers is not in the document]
klk1010: By putting it up here, it's serving as a safety check as well. The rest of the shift stuff will hit an error if `anchorRowIndex` is null and the behavior is different in the ctrl scenario, all though one could argue. This would only come into place for tables that allow no selections. In that case, one could argue that clicking control should also behave the same way - If there are no selections and you select the table using control, that row should become selected. I focused on shift because it seemed it would be clearly frustrating if users tried to click a range and nothing happened and an anchor row is required for that, but I can't think of a good reason not to support the same with control. I would just move that line above the `shiftKeyPressed` so it can still serve as a safety net.

src/components/Table.js
```diff
@@ -177,81 +197,231 @@ export default class Table extends React.Component<TableDefaultProps, TableProps
...
-        // if this is the only remaining selection and we don't allow empty selection, just don't update
-        return false;
-      } else {
-        // if it is already in the selection, remove it
-        selectedRowIds.splice(selectedRowIds.indexOf(rowId), 1);
-        rowId = selectedRowIds[selectedRowIds.length - 1];
-        this.props.onSelect(selectedRowIds, rowId);
-      }
-      return true;
+  getFormattedSelection = (selection): string => {
```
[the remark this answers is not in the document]
klk1010: Nope, forgot to delete it.

src/components/Table.js
klk1010: The <Table /> component should be usable as is without requiring a great deal of effort from whatever component renders it. `onSelect` is a convenience function. Using indices for selection tracking is far more performant because it's quicker to derive any value if you have the index, but if you just have the id, you have to search for the row, then derive what you want, instead of just grabbing it. However, now I know I can ignore backwards compatibility for MasterDetails and Table, it makes sense to keep these aligned and pass index through onSelect instead of id. It's trivial to derive id or row or whatever from the index. That will save the need to do the search in the render function of the MasterDetails component. It's unfortunate we're stuck with the inner-workings of react bootstrap table for now, it still requires an array of ids. I've worked on far more complex selection scenarios with larger data sets where performance was more important, and using index for selection in that scenario becomes a necessity. We don't know how this component will be used in the future, so we should follow best practices instead of using id which requires a search and find for every selection. When you have control of your own table, individual rows can determine if they are active or not, eliminating the need for some of these gymnastics that react-bootstrap requires, but for now, it's handy that it provides sorting and filtering out of the box for us in a way that appears to be meeting our needs for the moment.

It would also make these components more composable if the logic in `componentWillReceiveProps` in `MasterDetails` were moved into `<Table />` where `onSelect` would be called, if specified, so I'll make that change as well to make Table easier to use as a standalone.

klk1010: Also worth noting that our current version of react-bootstrap-table is deprecated. react-bootstrap-table2 exposes `rowIndex` in the `onSelect` function.

klk1010:
Ok, took another deep dive on this after being off it for a week. I feel strongly, it makes the most sense to:
Move the sorting and index assignment from the parent into the Table.
Track index internal with a set for optimal performance, to benefit from deduping, and Map lookup speed.
Provide the parent with the full data item in onSelect and not be opinionated about how it's used.

Indices aren't relevant outside of the Table, it's part of the internal tracking. The parent only cares about getting the relevant information in `onSelect` so it can get access to the selected rows. While it might seem like duplicate data tracking, it gets around the need to search for every connector by id over and over again, while not technically difficult to track just by id, it is expensive and non-performant. In `rowSelect` if I have the selected rows by index, determining what rows need to be selected next is much faster. row[index] has a lookup speed of O of 1. to find a row by id in the worst case is O of n where n is the total number of connectors. That might be ok if we needed to do that once, but we would need to do it for every single selected row. There are already necessary calculations that need to take place to determine fallback indices, etc. This one isn't necessary. If we want this table to support multiselect for more than a few items without lag, this is a better UX.

This also keeps the Table completely decoupled from MasterDetails. It could be used in any scenario with minimal extra effort - no more unintuitive `componentWillReceiveProps` in the parent, which really shouldn't care how the Table does its sorting.

PR #89, opened 2019-03-06
feat/PLAT-42668: Create a no label version of the ToggleSwitch

klk1010:
Non-breaking change.
Adds an optional `noLabel` prop to `<ToggleSwitch />. When defined as `true`, a rounded, material UI like toggle is used instead.
toggle-switch


src/components/ToggleSwitch.js
```diff
@@ -37,20 +41,93 @@ export default class ToggleSwitch extends React.Component<ToggleSwitchDefaultPro
...
     const containerClass = `toggle-switch-container ${disabledClass}`;
-    const onClass = `toggle-switch toggle-switch-on ${this.props.on ? 'selected' : ''} ${disabledClass}`;
-    const offClass = `toggle-switch toggle-switch-off ${this.props.on ? '' : 'selected'} ${disabledClass}`;
+    const onClass = `toggle-switch toggle-switch-on ${on ? 'selected' : ''} ${disabledClass}`;
+    const offClass = `toggle-switch toggle-switch-off ${on ? '' : 'selected'} ${disabledClass}`;
+
+    if (noLabel) {
+      const buttonStyle = on && !disabled
+        ? {
+          background: '#2F75B0',
```
benmvachon: Hard-coding color-codes like this is something we avoid in SUIT.
This toggle is the color of the default `@brand-primary` less variable in the `variables.less` file, but if somebody in Solutions or Professional Services creates a branded UI and changes this variable to suit a client or prospect, this toggle's color will not change with it.
I realize this isn't the ideal fix, but until we refactor the styling mechanism we should find existing class names which apply these colors and use those instead of calling the colors out directly by hash code.

klk1010: That makes sense, is there a specific file where all color variables are defined for suit?

documentation/style/custom.less
```diff
@@ -16,6 +16,14 @@
 @import "icons.data.svg.css";
 @import "temp.less";
 
+.attivio-primary-background {
```
luigivaldez: Are these styles in here twice (here and custom-general-overrides.less)?

klk1010: Yes, I forgot to remove them from the other file after adding them here.

klk1010:
Added examples of `<ToggleSwitch />` with `noLabel` to the styleguide docs.

PR #94, opened 2019-03-22
feat/PLAT-42929-D: Fix row update bug caused by multiselect refactor

klk1010:
Fixes row update issue caused by multiselect refactor.

### DIFF
--- a/src/components/MasterDetails.js
+++ b/src/components/MasterDetails.js
@@ -130,9 +130,9 @@ type MasterDetailsDefaultProps = {
 };
 
 type MasterDetailsState = {
-  /** The index for the most-recently selected row, to be displayed in the details pane. */
+  /** The data for the active row, to be displayed in the details pane. */
   detailsRow: {} | null;
-  /** The indices of the selected rows */
+  /** The data for the selected rows */
   selectedRows: Array<{}>;
 };
 
--- a/src/components/Table.js
+++ b/src/components/Table.js
@@ -215,7 +215,7 @@ export default class Table extends React.Component<TableDefaultProps, TableProps
 
   componentDidMount() {
     const { multiSelect, onSelect } = this.props;
-    const { activeRowIndex } = this.state;
+    const { sortedRows } = this.state;
 
     if (multiSelect) {
       document.addEventListener('keydown', this.keyDown);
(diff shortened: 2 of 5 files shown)


src/components/MasterDetails.js
```diff
@@ -130,9 +130,9 @@ type MasterDetailsDefaultProps = {
 };
 
 type MasterDetailsState = {
-  /** The index for the most-recently selected row, to be displayed in the details pane. */
+  /** The data for the active row, to be displayed in the details pane. */
```
klk1010: I forgot to update this comment previously. This PR doesn't change the behavior, but makes the jsDoc accurate.

src/components/Table.js
```diff
@@ -224,15 +224,24 @@ export default class Table extends React.Component<TableDefaultProps, TableProps
...
+      });
     }
   }
 
   componentWillReceiveProps(newProps: TableProps) {
     // If a rowComparator is specified, use it, otherwise, do a deep row comparison and reset selection if anything changed.
-    if ((newProps.rows && this.props.rows && newProps.rows.length !== this.props.rows.length)
-      || (newProps.rows && !this.props.rows)
-      || (!newProps.rows && this.props.rows)
+    if (newProps.rows.length !== this.props.rows.length
```
luigivaldez: I have a question about the way you're dealing with selections in this method: If a row "Foo" was previously selected and it was in the third position in the table (e.g., sortedRows[2]), and then the order of the rows changes so "Foo in at the 5th position (sortedRows[4]), won't it no longer be selected (and whatever is now at sortedRows[2] will be)? This would happen if the parent sorts the rows according to a different column, say... As a user, I think I would expect Foo to remain selected as I change the sort order (as long as it's still in the table).

src/components/Table.js
```diff
@@ -248,18 +257,28 @@ export default class Table extends React.Component<TableDefaultProps, TableProps
...
-      // If the parent provided an update hook, update the parent with the changes.
-      if (newProps.onSelect) {
-        const selectedRows = sortedRows.length > 0 ? [0] : [];
-        newProps.onSelect(selectedRows, activeRowIndex);
-      }
+    } else if (!isEqual(newProps.rows, this.props.rows)) {
+      // Selection hasn't changed, but other row data has update row data
+      const sortedRows = newProps.rows && newProps.rows.length > 0
+      ? newProps.rows.map((row, index) => {
+        return { ...row, index };
```
luigivaldez: Actually, another question: We should either document for the rows parameter that the objects cannot have a field named "index" or we should maybe rename the index field we're adding to the row objects to something more esoteric ('tableComponentIndex' or something?).

Antother caveat regarding the row objects is that the onSelect method is being passed the value from the sortedRows array and not the original object passed in. So if I pass an array of MyClass objects and I have a doSomething() method on MyClass, I won't be able to call row.doSomething() because the copy created with the spread operator won't have the prototype included, if I remember correctly.... I think we could probably just document this, too—as long as people are told explicitly, they can probably work around it.

src/components/Table.js
klk1010: When we discussed this with Dorit, it was my understanding that we decided that selection should be reset when sorting. filtering, search, or paging happens, if that's ok with you. I'd prefer to get this in for now. If you feel strongly about it we can raise it in the meeting today.

src/components/Table.js
klk1010: Because of the need for immutability in prop and state updates, more and more people are getting away from using classes for passing data around and instead using plain old js objects, especially with ES6. redux and functional programming practices have especially increased that mentality. That said, I agree I should give `index` a different name. Currently, the doc say a row must be an object, I'll update it to say it will be treated as an object.

luigivaldez: OK

luigivaldez:
Sorry, I had replied to your comments but didn't approve the whole PR. (I apparently had to hunt around to find the button again...)

Issue #9, opened 2018-05-04
Add maxresubmits param to searcher


raunak:
The 'maxresubmits' parameter allows AIE to resubmit AND queries as OR queries when AND query results in 0 results. 

By adding this param, we can control this functionality through configuration.js

EG Query: http://localhost:17000/admin/search?workflows=search&q=London+Paris&q.type=simple&q.maxresubmits=1

PR #95, opened 2019-03-25
searchui-issue-71: Add configurable Max Resubmits property to Searcher

brandon-bogan:
See https://github.com/attivio/searchui/issues/71 - essentially adding an additional configurable prop on Searcher that gets used for the maxResubmits rest param.


src/components/Searcher.js
```diff
@@ -440,6 +446,7 @@ class Searcher extends React.Component<SearcherDefaultProps, SearcherProps, Sear
       restParams.set('abc.enabled', ['true']);
       restParams.set('searchProfile', profiles);
     }
+    restParams.set('q.maxresubmits', [`${this.props.maxResubmits}`]);
```
luigivaldez: Question: is it true that this property should always be set? I.e., is there a use case for not setting it at all?

brandon-bogan:
@luigivaldez This should be up to date with attivio/master now

brandon-bogan:
FYI @luigivaldez this pull request will also fix Issue #9 from Raunak - https://github.com/attivio/suit/issues/9

PR #97, opened 2019-04-01
issue-96: Enable Autocomplete for Tags

brandon-bogan:
SearchResultTags now take an optional prop, `autoCompleteUri`. If not provided, tags behave as before with a basic input. If a uri is provided, the AutocompleteInput component will  be used instead. 

Accordingly, SearchResultTags now has functions to add a new tag based on an `Event` or a `string`. 

SearchResultTags also has a function for handling when users hit the Escape key, and this function is passed in as a newly added prop to AutoCompleteInput.


src/components/SearchResultTags.js
```diff
@@ -24,12 +26,27 @@ type SearchResultTagsProps = {
...
+   * configuration will be used.
+   * Otherwise, the search bar will not autocomplete.
+   * Note that this is relative to the baseUri field in the configuration.
+  */
+  autoCompleteUri: string,
+  /**
+   * Optional. The location of the node through which to interact with Attivio.
+   * Defaults to the value in the configuration.
+   */
+  baseUri: string,
```
klk1010: If `baseUri` is optional for the consumer of this component, this should be: `baseUri?: string`.

src/components/SearchResultTags.js
```diff
@@ -24,12 +26,27 @@ type SearchResultTagsProps = {
...
    */
   view360Label: string | null;
+  /**
+   * If set, the search bar’s input field will use autocomplete via this URI.
+   * Otherwise, if the configuration is available, the autoCompleteUri in the
+   * configuration will be used.
+   * Otherwise, the search bar will not autocomplete.
+   * Note that this is relative to the baseUri field in the configuration.
+  */
+  autoCompleteUri: string,
```
klk1010: If `autoCompleteUri` is optional for the consumer of this component, this should be: 
`autoCompleteUri?: string;`

klk1010:
Can you create examples for `AutoCompleteInput` in `documentation/components`?

src/components/SearchResultTags.js
brandon-bogan: Fixed in latest commit: https://github.com/brandon-bogan/suit/commit/9cfbb1d60f934586605da308f4e122b3766c5639

src/components/SearchResultTags.js
brandon-bogan: Fixed in latest commit: https://github.com/brandon-bogan/suit/commit/9cfbb1d60f934586605da308f4e122b3766c5639

klk1010:
Since `<AutoCompleteInput />` is already missing from the docs and isn't very meaningful without an active uri, I'm approving this without that change.

brandon-bogan:
> Early PR changes that already exist in master are showing as fresh changes in diff. There was a bad rebase/merge when changes from master were added that needs to be addressed in a clean rebase to keep the git history clean.

My bad - should be good now. I'm only seeing the changes related to this pull request coming through now.

PR #99, opened 2019-04-01
issue-98: Enable optimistic update for tags

brandon-bogan:
Search.js's updateRealTimeField function and Searcher's updateTags function now take onCompletion and onError callbacks, which get called once the update has been sent in but before the commit finishes. 

SearchResultTags creates the callbacks, and knows to show the update to the user and stop loading/spinning once one of these is called.

### DIFF
--- a/src/components/Searcher.js
+++ b/src/components/Searcher.js
@@ -719,8 +719,8 @@ class Searcher extends React.Component<SearcherDefaultProps, SearcherProps, Sear
   /**
    * Update the list of tags for the given document.
    */
-  updateTags(tags: Array<string>, docId: string): Promise<any> {
-    return this.search.updateRealtimeField(docId, FieldNames.TAGS, tags);
+  updateTags(tags: Array<string>, docId: string, onCompletion: () => void, onError: (error: string) => void): Promise<any> {
+    return this.search.updateRealtimeField(docId, FieldNames.TAGS, tags, onCompletion, onError);
   }
 
   props: SearcherProps;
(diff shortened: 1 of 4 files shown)


brandon-bogan:
> Hey Brandon, can you make these changes on your tag_autocomplete branch instead. This way, you've created a second pull request instead of just augmenting the first one and it's not really possible to reconcile them both together... (If you commit changes to a branch that's already the source of a pull request, the changes are added to the PR automatically—GitHub can even let each of us see the changes since the previous review...)
> Thanks!!

Done! See https://github.com/brandon-bogan/suit/commit/8a4810b18276cff50a2812823f24944525789fbf for the merge into the `tag_autocomplete` branch

src/components/Searcher.js
```diff
@@ -719,8 +719,8 @@ class Searcher extends React.Component<SearcherDefaultProps, SearcherProps, Sear
   /**
    * Update the list of tags for the given document.
    */
-  updateTags(tags: Array<string>, docId: string): Promise<any> {
-    return this.search.updateRealtimeField(docId, FieldNames.TAGS, tags);
+  updateTags(tags: Array<string>, docId: string, onCompletion: () => void, onError: (error: string) => void): Promise<any> {
```
klk1010: I might be missing something, but I don't see a place where `updateTags` is used. This change isn't harmful, but I'd expect to see any place using `updateTags to be updated as well.

klk1010: nm, I see it being used through context and you already updated that.

klk1010:
If this is being included in PR#96, can this PR be closed?

PR #100, opened 2019-04-08
Feat/plat 42929 a

klk1010:
Introduces option to disable the confirmation button in `<ConfirmationDialog />`.

This is useful when the action being confirmed takes a while. It allows displaying a spinner, for example, and disabling the confirm button while the spinner is active.


klk1010:
The use case for connector admin is that bulk actions can take a while to complete - several seconds. While waiting for completion, I don't want the dialog box to close, so I add a pending/loading state and a spinner with text indicating the action. While the requested action is completing, I'd like to disable the confirmation button to prevent duplicate requests and also indicate that an additional request isn't possible until the original request completes. Once the loading/pending state completes, the dialog closes.

PR #103, opened 2019-04-11
Feat/plat 42929 e

klk1010:
Addresses issue in `<Table />` where the `selectedRow` value passed through `onSelect` was always `null`. The result was that after row data changed, any consumer would lack the data for the active Row. `<MasterDetails />`, for example would have a `null` `detailRow`. 

The impact in connector admin: This manifested as an empty detail pane after a search was performed. This fixes that issue.


src/components/Table.js
```diff
@@ -259,8 +259,8 @@ export default class Table extends React.Component<TableDefaultProps, TableProps
 
       // If the parent provided an update hook, update the parent with the changes.
       if (newProps.onSelect) {
-        const selectedRows = activeRowIndex ? [sortedRows[activeRowIndex]] : [];
-        const selectedRow = activeRowIndex ? sortedRows[activeRowIndex] : null;
+        const selectedRows = activeRowIndex !== null && newProps.noEmptySelection ? [sortedRows[activeRowIndex]] : [];
+        const selectedRow = activeRowIndex !== null && newProps.noEmptySelection ? sortedRows[activeRowIndex] : null;
```
benmvachon: what if `activeRowIndex` doesn't equal `null` but `newProps.noEmptySelection` is `false`?

klk1010: If `noEmptySelection` is false, then instead of setting the selection to the first row, there should be no selection.

However, I need to add the same condition to lines 256 - 258, otherwise that case will have the same bug it does now because the table will still show the row as selected.

Now:
```
const selectedRowIndices = sortedRows.length > 0 ? new Set([0]) : new Set([]);
const anchorRowIndex = sortedRows.length > 0 ? 0 : null;
const activeRowIndex = sortedRows.length > 0 ? 0 : null;
```

Should be: 
```
const selectedRowIndices = sortedRows.length > 0 && noEmptySelection
  ? new Set([0])
  : new Set([]);
const anchorRowIndex = sortedRows.length > 0 && noEmptySelection
  ? 0
  : null;
const activeRowIndex = sortedRows.length > 0 && noEmptySelection
  ? 0
  : null;
```

klk1010: Same for `componentDidMount`. I'm doing that in the constructor already.

klk1010: updated

benmvachon: If `activeRowIndex` is set to say 5 and `newProps.noEmptySelection` is false, then we'll send
```
newProps.onSelect([], null);
```
Is that what we want to do? Shouldn't we maintain the active row as 5?

klk1010: `const activeRowIndex = sortedRows.length > 0 ? 0 : null;`
`activeRowIndex` can only be `0` or `null` in this case. This scenario is only triggered when the number of connectors changes or the rowComparator detects a change. For connector admin, that would mean a row id changed. This is basically resetting the selected row to the first in the table, when `noEmptySelection` is true or resetting the table to no selection if `noEmptySelection` is false. This was already happening in the constructor, so that the initial selection follows the same logic. If `noEmptySelection` is true, the first row is selected. If `noEmptySelection` is false, no row is selected. If general data on a row changes - anything but the id, then the selection does not change. The reason this is necessary is because a search or filter change could result in the same number of connectors being displayed. In that case, we need to update the selection because a different connector could take the place of the same index position. During product discussions, we decided it was better to reset the selection on search, filter, and pagination. In some cases the previous selection may not exist. We could discuss doing something different - updating the selection to whatever item holds the current index, and if the index is greater than the table size (ie. the number of items in the table decreased and the previous selection was at the end), we could instead update the selection to the last item in the table. You could raise this in our next meeting if you think it would be better. The change in this PR enforces the same selection behavior that happens on initial load.

PR #104, opened 2019-04-15
Feature/PLAT-42690 fix Table selection behavior on rows update

benmvachon:
I basically just renamed all the `selectedRowIndices` variables in `<Table />`
to `selectedIndices` to avoid the confusion that I think ultimately led to
`selectedRowIndices` (which isn't actually defined in `TableState` or used in
the component) being updated in the state when it should have been
`selectedIndices`.

Motivation:
In connectoradmin, when the user changes pages or filters the list, the state
of the `selectedIndices` is not updated properly, so the active row is not
highlighted properly in the table.

Changes:
* In the Table's `componentDidMount` hook, we set the `selectedIndices` state
according to the same logic that initializes it in the constructor.
* `selectedRowIndices` variables were renamed to `selectedIndices` to reflect
`TableState` type definition.
* `cleanFlow` script was renamed to `cleanflow` to match the `package.json`/other
script names

Testing:
* manual testing of UI
* ran flow
* ran lint
* ran validatecomponents
* ran fullbuild


src/components/Table.js
```diff
@@ -234,7 +234,7 @@ export default class Table extends React.Component<TableDefaultProps, TableProps
       this.setState({
         activeRowIndex,
         anchorRowIndex: activeRowIndex,
-        selectedRowIndices: sortedRows.length > 0 ? new Set(selectedRows) : new Set(),
+        selectedIndices: noEmptySelection ? new Set([0]) : new Set([]),
```
luigivaldez: What's the deal with this line? Don't you want to only set selectedIndices to `new Set[0]` if there are actually rows in the table?

benmvachon: If there are no rows in the table, then having `selectedIndices` set to `new Set([0])` should have no effect.
I think it's safe to do this. It's the way `selectedIndices` is initialized in the constructor on line 209 above.

luigivaldez: OK

PR #105, opened 2019-04-17
PLAT-42690 fix versions in html files

benmvachon:
This was necessary for me to be able to run `publishlib`.


benmvachon:
It was changed automatically when I did a `publishlib`, but then the `publishlib` failed because I had uncommitted changes.

PR #110, opened 2019-04-23
issue-108: Add saved search feature

nikhilj13:
# Save searches

Save search feature allows you to save your search (including any filters that are applied) which can later be revisited by clicking on the saved search title. This feature is disabled by default.

## How to enable save search
You can turn on the save search option by passing in an additional prop to SearchBar component, this prop is called _allowSavedSearch_. For example: 
```
<SearchBar allowSavedSearch />
```
This would show a heart icon with a drop down next to the Search Bar, which means the UI component for saved search is enabled.
> **Note:** In order to make this feature work, we need to create a new zone to store the saved searches. This feature will **NOT** work if you haven't configured and specified a new zone for saved searches.

## Configuring a new zone to store the saved searches
*Follow these steps to configure a new zone for storing the saved searches:*

 1. Zones are configured via the [Index Feature](https://answers.attivio.com/display/extranet52/Configure+the+Index), in **[project-dir]\conf\features\core\Index.index.xml**.  It should look similar to:
	 ```
	 <?xml version="1.0" encoding="UTF-8"?>

	<ff:features xmlns:ff="http://www.attivio.com/configuration/config" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:fbase="http://www.attivio.com/configuration/features/base" xmlns:f="http://www.attivio.com/configuration/features/core" xsi:schemaLocation="http://www.attivio.com/configuration/config http://www.attivio.com/configuration/config.xsd http://www.attivio.com/configuration/features/base http://www.attivio.com/configuration/features/baseFeatures.xsd http://www.attivio.com/configuration/features/core http://www.attivio.com/configuration/features/coreFeatures.xsd">
	  <f:index enabled="true" name="index" profile="true">
	    <f:partitionSet size="1"/>
	    <f:writer logCommits="true" nodeset="ingestion" search="true"/>
	  </f:index>
	</ff:features>
	```

 2. Edit this file to add in a new index zone. For example, we're creating a new index zone called *testZone* in the example below:
	 ```
	 <?xml version="1.0" encoding="UTF-8"?>

	<ff:features xmlns:ff="http://www.attivio.com/configuration/config" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:fbase="http://www.attivio.com/configuration/features/base" xmlns:f="http://www.attivio.com/configuration/features/core" xsi:schemaLocation="http://www.attivio.com/configuration/config http://www.attivio.com/configuration/config.xsd http://www.attivio.com/configuration/features/base http://www.attivio.com/configuration/features/baseFeatures.xsd http://www.attivio.com/configuration/features/core http://www.attivio.com/configuration/features/coreFeatures.xsd">
		<f:index enabled="true" name="index" profile="true">
			<f:writer defaultZoneName="default" logCommits="true" nodeset="ingestion" search="true">
				<f:zone name="default" />
				<f:zone name="testZone" />
			</f:writer>
		</f:index>
	</ff:features>
	```
> If you don't configure any zones, there is always one zone named _default_. However, if you configure zones, you must explicitly configure the default zone in addition to the other zone. Additionally, you can also set up separate search workflows for these zones so you don't get the results from your *testZone* as a part of the search results.

3. Configure the save search in SearchUI's configuration.properties file. You can specify the name of the new index zone for saved searches, the table name for all the documents for saved searches and the search workflow to be used for querying the mentioned zone. Below is an example with default values:
	```
	SearchBar: {
		// configures the search workflow used for saved searches
		savedSearchWorkflow: 'searchSavedSearches',
		// configures the table name for saved searches
		savedSearchTable: 'savedSearches',
		// specifies the index zone for saved searches
		savedSearchZone: 'savedSearches',
	}
	```

## How to use saved searches
After you enable and configure saved searches, you should have this component right next to the search bar:
SavedSearch

### Saving a search
You can save searches by clicking on the first option in the drop-down menu 'SAVE SEARCH'. It will open up a modal that asks for the name for this search. 
Save a search
> If you don't enter a name, the search term is used as the name.

Click on save search button to save the search.

### Revisiting a search
You can revisit a search by clicking on the search title in the drop-down for saved search.
Revisit a search

### Deleting a saved search
You can also delete a saved search, there's a **x** next to every saved search in the drop-down. Simply click on the **x** to delete a saved search.
Delete a search
> Note that the drop down only shows the 10 most recent saved searches.

### DIFF
--- a/src/components/SearchBar.js
+++ b/src/components/SearchBar.js
@@ -3,17 +3,27 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import { withRouter } from 'react-router-dom';
 
-import Dropdown from 'react-bootstrap/lib/Dropdown';
-import Glyphicon from 'react-bootstrap/lib/Glyphicon';
-import MenuItem from 'react-bootstrap/lib/MenuItem';
+import {
+  Dropdown,
+  Glyphicon,
+  MenuItem,
+  Button,
+  Modal,
+  Row,
+  Col,
+} from 'react-bootstrap';
 
 import Configurable from './Configurable';
 import AutoCompleteInput from './AutoCompleteInput';
+import AuthUtils from '../util/AuthUtils';
+import SimpleQueryRequest from '../api/SimpleQueryRequest';
+import QueryResponse from '../api/QueryResponse';
 
 declare var webkitSpeechRecognition: any; // Prevent complaints about this not existing
 
 type SearchBarProps = {
   history: PropTypes.object.isRequired;
+  location: PropTypes.object.isRequired;
   /** If set, this will be styled to live inside a Masthead component. */
   inMasthead: boolean;
   /**
@@ -54,6 +64,14 @@ type SearchBarProps = {
   buttonLabel: string;
   /** If set, this is the route to navigate to upon executing a search. By default, no navigation will occur when searching. */
   route: string | null;
+  /** If set, the save search option is displayed next to the SearchBar, false by default */
+  allowSavedSearch: boolean;
+  /** Optional, specifies the workflow that should be used for querying the index which stores saved searches */
+  savedSearchWorkflow?: string;
+  /** Optional, specifies the table name assigned to the docs for saved searches, defaults to 'savedSearches' */
+  savedSearchTable: string;
+  /** Specifies the zone name for the zone that stores saved searches, defaults to 'savedSearches' */
+  savedSearchZone: string;
 };
 
 type SearchBarDefaultProps = {
@@ -66,11 +84,19 @@ type SearchBarDefaultProps = {
   autoCompleteUri: string | null;
   route: string | null;
   baseUri: string;
+  allowSavedSearch: boolean;
+  savedSearchTable: string,
+  savedSearchZone: string;
 };
 
 type SearchBarState = {
   recognizing: boolean;
   suggestions: Array<string>;
+  response?: QueryResponse;
+  error?: string;
+  showSaveSearchModal: boolean;
+  savedSearchTitle: string;
+  savedSearchList: Array<any>;
 };
 
 /**
@@ -88,6 +114,10 @@ class SearchBar extends React.Component<SearchBarDefaultProps, SearchBarProps, S
     autoCompleteUri: null,
     route: null,
     baseUri: '',
+    allowSavedSearch: false,
+    savedSearchWorkflow: '',
+    savedSearchTable: 'savedSearches',
+    savedSearchZone: 'savedSearches',
   };
 
   static contextTypes = {
@@ -104,7 +134,11 @@ class SearchBar extends React.Component<SearchBarDefaultProps, SearchBarProps, S
       query: '',
       recognizing: false,
       suggestions: [],
-
+      response: undefined,
+      error: undefined,
+      showSaveSearchModal: false,
+      savedSearchTitle: '',
+      savedSearchList: [],
     };
     (this: any).doKeyPress = this.doKeyPress.bind(this);
     (this: any).doSearch = this.doSearch.bind(this);
@@ -112,12 +146,23 @@ class SearchBar extends React.Component<SearchBarDefaultProps, SearchBarProps, S
     (this: any).queryChanged = this.queryChanged.bind(this);
     (this: any).updateQuery = this.updateQuery.bind(this);
     (this: any).languageChanged = this.languageChanged.bind(this);
+    (this: any).applySavedSearch = this.applySavedSearch.bind(this);
+    (this: any).showSaveSearchModal = this.showSaveSearchModal.bind(this);
+    (this: any).hideSaveSearchModal = this.hideSaveSearchModal.bind(this);
+    (this: any).getSavedSearches = this.getSavedSearches.bind(this);
+    (this: any).searchTitleEntered = this.searchTitleEntered.bind(this);
     if (this.props.allowVoice && !('webkitSpeechRecognition' in window)) {
       console.log('Requested speech recognition but the browser doesn’t support it'); // eslint-disable-line no-console
     }
   }
 
   state: SearchBarState;
+
+  componentWillMount() {
+    if (this.props.allowSavedSearch) {
+      this.getSavedSearches();
+    }
+  }
 
   getSuggestionList() {
     if (!this.state.suggestions || this.state.suggestions.length === 0) {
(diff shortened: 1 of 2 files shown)


klk1010:
@nikhilj13, the notes on this PR are fantastic. Can you create an answer page and copy this info there? I'd hate to lose track of this functionality.

src/components/SearchBar.js
```diff
@@ -3,17 +3,27 @@ import React from 'react';
...
 import AutoCompleteInput from './AutoCompleteInput';
+import AuthUtils from '../util/AuthUtils';
+import SimpleQueryRequest from '../api/SimpleQueryRequest';
+import QueryResponse from '../api/QueryResponse';
 
 declare var webkitSpeechRecognition: any; // Prevent complaints about this not existing
 
 type SearchBarProps = {
   history: PropTypes.object.isRequired;
+  location: PropTypes.object.isRequired;
```
klk1010: If we make the `location` required this would become a breaking change. Can you refactor things a bit to make this optional?

src/components/SearchBar.js
```diff
@@ -54,6 +64,14 @@ type SearchBarProps = {
   buttonLabel: string;
   /** If set, this is the route to navigate to upon executing a search. By default, no navigation will occur when searching. */
   route: string | null;
+  /** If set, the save search option is displayed next to the SearchBar, false by default */
+  allowSavedSearch: boolean;
+  /** Optional, specifies the workflow that should be used for querying the index which stores saved searches */
+  savedSearchWorkflow?: string;
+  /** Optional, specifies the table name assigned to the docs for saved searches, defaults to 'savedSearches' */
+  savedSearchTable: string;
```
klk1010: Indicate this is optional in flow:
`savedSearchTable?: string;`

src/components/SearchBar.js
```diff
@@ -54,6 +64,14 @@ type SearchBarProps = {
...
+  /** Specifies the zone name for the zone that stores saved searches, defaults to 'savedSearches' */
+  savedSearchZone: string;
```
klk1010: Is this optional?
`savedSearchZone?: string;`

src/components/SearchBar.js
```diff
@@ -133,6 +178,86 @@ class SearchBar extends React.Component<SearchBarDefaultProps, SearchBarProps, S
...
+    const { location } = this.props;
+    this.props.history.push({
+      pathname: location.pathname,
+      search: query,
+    });
+  }
+
+  // populates the saved searches in the saved searches drop down
+  populateSavedSearches() {
+    const response = this.state.response;
```
klk1010: Prefer desctructuring.
`const { response } = this.state;`

src/components/SearchBar.js
```diff
@@ -133,6 +178,86 @@ class SearchBar extends React.Component<SearchBarDefaultProps, SearchBarProps, S
...
+      qr.restParams.set('zones', [`${savedSearchZone}`]);
+      if (savedSearchWorkflow) {
+        qr.workflow = savedSearchWorkflow;
+      }
+      qr.queryLanguage = 'advanced';
+      searcher.doCustomSearch(qr, (response: QueryResponse | null, error: string | null) => {
+        if (response) {
+          this.setState(
+            {
+              response,
```
klk1010: Could you create a class for the response in the `api directory` and move the formatting you do later on into it to be used here?

src/components/SearchBar.js
```diff
@@ -66,11 +84,19 @@ type SearchBarDefaultProps = {
   autoCompleteUri: string | null;
   route: string | null;
   baseUri: string;
+  allowSavedSearch: boolean;
```
klk1010: New props should be optional for backwards compatibility.
`allowSavedSearch?: boolean;`

src/components/SearchBar.js
```diff
@@ -54,6 +64,14 @@ type SearchBarProps = {
   buttonLabel: string;
   /** If set, this is the route to navigate to upon executing a search. By default, no navigation will occur when searching. */
   route: string | null;
+  /** If set, the save search option is displayed next to the SearchBar, false by default */
+  allowSavedSearch: boolean;
```
klk1010: New props should be optional for backwards compatibility.
`allowSavedSearch?: boolean;`

src/components/SearchBar.js
```diff
@@ -133,6 +178,86 @@ class SearchBar extends React.Component<SearchBarDefaultProps, SearchBarProps, S
...
+    if (response) {
+      const menuItemList = [];
+      response.documents.forEach((doc) => {
+        const savedSearch = {};
+        const htmlString = doc.getFirstValue('title');
+        savedSearch.id = doc.getFirstValue('.id');
+        savedSearch.title = htmlString.replace(/<[^>]+>/g, '');
+        savedSearch.query = doc.getFirstValue('query_s');
+        savedSearch.queryString = doc.getFirstValue('query_string_s');
+        let savedSearchTitle = savedSearch.title;
```
klk1010:
```
const savedSearchTitle = savedSearch.title > 15 ? `${savedSearch.title.substring(0, 15)}...` : savedSearch.title;
```
[`substring()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substring) does not modify the original string, it returns a brand new string, so it's safe to use the original string.

src/components/SearchBar.js
```diff
@@ -133,6 +178,86 @@ class SearchBar extends React.Component<SearchBarDefaultProps, SearchBarProps, S
...
+                float: 'right',
+                fontWeight: 'bold',
+              }}
+            >
+              &times;
+            </span>
+          </MenuItem>,
+        );
+      });
+      this.setState({ savedSearchList: menuItemList });
```
klk1010: Storing jsx as state is generally considered bad practice. It's usually an indicator that you need to refactor your component, break out a function, or create a building block component.
https://reactjs.org/docs/thinking-in-react.html#step-3-identify-the-minimal-but-complete-representation-of-ui-state

src/components/SearchBar.js
```diff
@@ -223,6 +357,44 @@ class SearchBar extends React.Component<SearchBarDefaultProps, SearchBarProps, S
     }
   }
 
+  // saves the current search from the url and feed it to a separate index zone
+  saveThisSearch(searcherState) {
+    const searcher = this.context.searcher;
```
klk1010: Prefer destructuring:
`const { searcher } = this.context;`

src/components/SearchBar.js
```diff
@@ -223,6 +357,44 @@ class SearchBar extends React.Component<SearchBarDefaultProps, SearchBarProps, S
...
+    const { savedSearchZone, savedSearchTable } = this.props;
+    const currentSearch = searcher.generateLocationQueryStringFromState(searcherState);
+    const username = AuthUtils.getLoggedInUserId();
+    const loggedDateTime = new Date().toISOString();
+    const id = username.concat(loggedDateTime);
+    const title = this.state.savedSearchTitle;
```
klk1010: Place this at the top of the function somewhere above `line 364`. Prefer destructuring:
`const { savedSearchTitle: title } = this.state;`

src/components/SearchBar.js
```diff
@@ -223,6 +357,44 @@ class SearchBar extends React.Component<SearchBarDefaultProps, SearchBarProps, S
...
+    const queryString = searcher.state.query;
+
+    const jsonDoc = '{ "fields" : { "title" : '
+      + `[ "${title}" ], "query_s" : [ "${currentSearch}" ], "query_string_s" : [ "${queryString}" ], `
+      + `"username_s" : [ "${username}" ], "date" : [ "${loggedDateTime}" ], "table" : [ "${savedSearchTable}" ] }, `
+      + `"id" : "${id}", "zone" : "${savedSearchZone}" }`;
+    searcher.search.updateSavedSearches(JSON.parse(jsonDoc), this.getSavedSearches);
```
klk1010: Response parsing should be done in a class.

src/components/SearchBar.js
```diff
@@ -348,9 +521,127 @@ class SearchBar extends React.Component<SearchBarDefaultProps, SearchBarProps, S
...
+            color: '#5d7c89',
+            cursor: 'default',
+          }}
+        >
+          No Saved Searches
+        </span>
+      </MenuItem>
+    );
+
+    const saveQuery = this.props.allowSavedSearch ? (
```
klk1010: Move this to out's own render function. Same with the other saved variables.

klk1010 (changes requested):
Address comments.

PR #114, opened 2019-04-24
issue-113: Add share search

nikhilj13:
Adds a share search icon next to the SearchBar. This icon is hidden by default but if you set the _sharesearch_ prop for SearchBar to true, you'll be able to see it.

Here's an example:
ShareSearch


src/components/SearchBar.js
```diff
@@ -283,6 +288,18 @@ class SearchBar extends React.Component<SearchBarDefaultProps, SearchBarProps, S
     }
   }
 
+  //eslint-disable-next-line
```
luigivaldez: Why is the eslint-disable-next-line here?

src/components/SearchBar.js
```diff
@@ -283,6 +288,18 @@ class SearchBar extends React.Component<SearchBarDefaultProps, SearchBarProps, S
...
+  shareSearch() {
+    const username = AuthUtils.getUserName(AuthUtils.getSavedUser());
+    const yourMessage =
+      'Hey,\n\nI think you would be interested in these search results that I found using the most cognitive and ' +
```
luigivaldez: This message should be a property. In a PoV this makes sense, but if someone wants to use it in a production system, they'd probably want to have different text.

src/components/SearchBar.js
```diff
@@ -283,6 +288,18 @@ class SearchBar extends React.Component<SearchBarDefaultProps, SearchBarProps, S
...
+      `intuitive search platform called Attivio. Here is the link :)\n${window.location.href}
+      \nLet me know if you have any questions!\n\nBest,\n${username}`;
+    const subject = 'Search results I found using Attivio!';
+    const emailLink = 'email@example.com';
```
luigivaldez: The target email should be a property too, although it should default to being empty, I think... In general, it doesn't make sense to have the address come up pre-filled to something no one will use... (for a PoV, sure...)

src/components/SearchBar.js
```diff
@@ -400,9 +418,29 @@ class SearchBar extends React.Component<SearchBarDefaultProps, SearchBarProps, S
         />
       );
 
+    const shareSearch = this.props.shareSearch ? (
+      <span
+        className="attivio-smalltoolbar-btn"
+        title="Share this search via Email"
```
luigivaldez: "Email" shouldn't be capitalized.

luigivaldez (changes requested):
I think it makes more sense to have this be a completely separate component from the SearchBar. On the search page (in Search UI), you can then include it right after the `<SearchBar>` component. but this way you don't have to complicate SearchBar and you don't have to pass all of the properties you want this component to have through the SearchBar's properties.
Does that make sense?

PR #115, opened 2019-04-24
Add SwipeViews Component

nikhilj13:
Adds a new component called SwipeViews which uses the react-swipe library to show multiple views that you can navigated through by using the left and right buttons.

Below is an example:
SwipeViews

### DIFF
--- a/src/components/SwipeViews.js
+++ b/src/components/SwipeViews.js
@@ -1,0 +1,53 @@
+/* eslint-disable react/no-string-refs */
+import React from 'react';
+import ReactSwipe from 'react-swipe';
+import { Row, Col, Glyphicon } from 'react-bootstrap';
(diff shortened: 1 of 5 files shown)


src/components/SwipeViews.js
```diff
@@ -0,0 +1,56 @@
...
+  next() {
+    this.refs.reactSwipe.next();
+  }
+
+  prev() {
+    this.refs.reactSwipe.prev();
+  }
+
+  render() {
+    const swipeOptions = {
```
luigivaldez: Do we want to expose any of the values in `swipeOptions` as properties of the component?

nikhilj13: Done, all of the options are configurable now.

PR #116, opened 2019-04-25
Add FacetTabs Component

nikhilj13:
Adds a FacetTabs component that shows facet as a tab. The main use for facet tabs would be to show the search results based on the data source. Below is an example:

facetTabs

Also modifies FacetResults to take in a list of facets that should be hidden. If we're showing a facet as tabs we most likely would not want the facet to show up in facet results. It is configurable from SearchUI's configuration properties file. Here's an example:
```
FacetResults: {
	// facets that should be hidden from the FacetResults
	hideFacet: [ 'table' ],
  },
```

### DIFF
--- a/src/components/FacetResults.js
+++ b/src/components/FacetResults.js
@@ -4,6 +4,7 @@ import PropTypes from 'prop-types';
 
 import Facet from './Facet';
 import SearchFacet from '../api/SearchFacet';
+import Configurable from './Configurable';
 
 type FacetResultsProps = {
   /** The facet field names that should be displayed as pie charts */
@@ -42,6 +43,10 @@ type FacetResultsProps = {
    * buckets. By default, facets with no buckets will be hidden.
    */
   showEmptyFacets: boolean;
+  /**
+   * An optional list of facets that should not be shown but are needed for other reasons
+   */
+   hideFacet: Array<string>;
 };
 
 type FacetResultsDefaultProps = {
@@ -57,6 +62,7 @@ type FacetResultsDefaultProps = {
   orderHint: Array<string>;
   entityColors: Map<string, string>;
   showEmptyFacets: boolean;
+  hideFacet: Array<string>,
 };
 
 /**
@@ -67,7 +73,7 @@ type FacetResultsDefaultProps = {
  * not covered by one of these property's lists will be displayed
  * in a standard "More…" list.
  */
-export default class FacetResults extends React.Component<FacetResultsDefaultProps, FacetResultsProps, void> {
+class FacetResults extends React.Component<FacetResultsDefaultProps, FacetResultsProps, void> {
   static defaultProps = {
     pieChartFacets: null,
     barChartFacets: null,
@@ -81,6 +87,7 @@ export default class FacetResults extends React.Component<FacetResultsDefaultPro
     orderHint: [],
     entityColors: new Map(),
     showEmptyFacets: false,
+    hideFacet: [],
   };
 
   static contextTypes = {
@@ -129,6 +136,11 @@ export default class FacetResults extends React.Component<FacetResultsDefaultPro
   }
 
   shouldShow(facet: SearchFacet): boolean {
+    if (this.props.hideFacet && this.props.hideFacet.length > 0) {
+      if (this.props.hideFacet.includes(facet.field)) {
+        return false;
+      }
+    }
     if (this.props.showEmptyFacets || (facet && facet.buckets && facet.buckets.length > 0)) {
       return true;
     }
@@ -186,3 +198,5 @@ export default class FacetResults extends React.Component<FacetResultsDefaultPro
     );
   }
 }
+
+export default Configurable(FacetResults);
--- a/src/components/FacetTabs.js
+++ b/src/components/FacetTabs.js
@@ -1,0 +1,153 @@
+// @flow
+
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import Configurable from './Configurable';
+import SearchFacet from '../api/SearchFacet';
+import SearchFacetBucket from '../api/SearchFacetBucket';
+import BusyIndicator from './BusyIndicator';
+
+type FacetTabsProps = {
+  /**
+   * Name of the facet that should be displayed as Tabs, default is 'table'
+   */
+  facetName: string; //eslint-disable-line
+  /**
+   * Optional callback to be used when a tab is clicked instead of default (which applies the facet filter)
+   */
+  handleClick?: () => void;
+  /**
+   * Optional facet that should be used instead of facetName
+   */
+  facet?: SearchFacet | null; //eslint-disable-line
+  /**
+   * The color that should be used for the background of facet tabs
+   */
+  backgroundColor?: string;
+};
+
+type FacetTabsState = {
+  tabsList: Array<any>;
+};
+
+class FacetTabs extends React.Component<FacetTabsProps, FacetTabsProps, FacetTabsState> {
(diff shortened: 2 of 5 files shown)


src/components/FacetResults.js
```diff
@@ -42,6 +43,10 @@ type FacetResultsProps = {
    * buckets. By default, facets with no buckets will be hidden.
    */
   showEmptyFacets: boolean;
+  /**
+   * An optional list of facets that should not be shown but are needed for other reasons
+   */
+   hideFacet: Array<string>;
```
klk1010: Add `?` to indicate this is optional. Make the prop name plural to indicate it's an array, not a boolean.
```suggestion
   hiddenFacets?: Array<string>;
```

src/components/FacetResults.js
```diff
@@ -57,6 +62,7 @@ type FacetResultsDefaultProps = {
   orderHint: Array<string>;
   entityColors: Map<string, string>;
   showEmptyFacets: boolean;
+  hideFacet: Array<string>,
```
klk1010: Rename prop to indicate this is an array not a boolean.
```suggestion
  hiddenFacets: Array<string>,
```

src/components/FacetResults.js
```diff
@@ -81,6 +87,7 @@ export default class FacetResults extends React.Component<FacetResultsDefaultPro
     orderHint: [],
     entityColors: new Map(),
     showEmptyFacets: false,
+    hideFacet: [],
```
klk1010:
```suggestion
    hiddenFacets: [],
```

src/components/FacetResults.js
```diff
@@ -129,6 +136,11 @@ export default class FacetResults extends React.Component<FacetResultsDefaultPro
   }
 
   shouldShow(facet: SearchFacet): boolean {
+    if (this.props.hideFacet && this.props.hideFacet.length > 0) {
```
klk1010:
```suggestion
    if (this.props.hiddenFacets && this.props.hiddenFacets.length > 0) {
```

src/components/FacetResults.js
```diff
@@ -129,6 +136,11 @@ export default class FacetResults extends React.Component<FacetResultsDefaultPro
...
+      if (this.props.hideFacet.includes(facet.field)) {
```
klk1010:
```suggestion
      if (this.props.hiddenFacets.includes(facet.field)) {
```

src/components/FacetTabs.js
```diff
@@ -0,0 +1,153 @@
...
+    this.state = {
+      tabsList: [],
+    };
+
+    (this: any).getTabsFromFacet = this.getTabsFromFacet.bind(this);
+    (this: any).populateTabs = this.populateTabs.bind(this);
+    (this: any).handleTabClick = this.handleTabClick.bind(this);
+  }
+
+  componentWillMount() {
```
klk1010: Avoid using `componentWillMount`

src/components/FacetTabs.js
```diff
@@ -0,0 +1,153 @@
...
+            }
+          });
+        }
+      }
+    }
+  }
+
+  render() {
+    const tabData = this.state.tabsList;
+    const tabs =
```
klk1010: Pull `tabs` out into `renderTabs`

src/components/FacetTabs.js
```diff
@@ -0,0 +1,153 @@
...
+            className="facet-tab"
+          >
+            <div className="facet-tab-text">
+              {value} <span style={{ color: 'gray' }}>({count})</span>
+            </div>
+          </div>,
+        );
+      }
+    });
+    this.setState({ tabsList: tabs });
```
klk1010: From react docs, ("Thinking in React")[https://reactjs.org/docs/thinking-in-react.html#step-3-identify-the-minimal-but-complete-representation-of-ui-state]
>Ask three questions about each piece of data:

>Is it passed in from a parent via props? If so, it probably isn’t state.
Does it remain unchanged over time? If so, it probably isn’t state.
_Can you compute it based on any other state or props in your component? If so, it isn’t state._

With that in mind, if you're storing html as state you should reconsider.

src/components/FacetTabs.js
```diff
@@ -0,0 +1,153 @@
...
+    this.populateTabs(this.props);
+  }
+
+  componentWillReceiveProps(newProps: FacetTabsProps) {
+    this.populateTabs(newProps);
+  }
+
+  //eslint-disable-next-line
+  numberWithCommas(x) {
+    return x.toString().replace(/\B(?=(\d{3})+(?!\d))/g, ',');
```
luigivaldez: You can just use x.toLocaleString() to do this without needing a special function.

`47626.toLocaleString()` gives 47,626.

See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/toLocaleString

src/components/FacetTabs.js
```diff
@@ -0,0 +1,153 @@
...
+  }
+
+  handleTabClick(bucket: SearchFacetBucket, label: string) {
+    this.context.searcher.addFacetFilter(label, bucket.value, bucket.filter);
+  }
+
+  getTabsFromFacet(facet: SearchFacet) {
+    const tabs = [];
+    facet.buckets.forEach((bucket: SearchFacetBucket) => {
```
luigivaldez: Do you need to have a limit on the number of tabs? What if there are 70 buckets?

src/components/FacetTabs.js
```diff
@@ -0,0 +1,153 @@
...
+        );
+      }
+    });
+    this.setState({ tabsList: tabs });
+  }
+
+  populateTabs(props: FacetTabsProps) {
+    const { facetName, facet } = props;
+
+    if (facet && facet.buckets) {
```
luigivaldez: What happens if the facet had buckets before but doesn't any more (e.g. because of a different search term)? Will the old tabs not be replaced?

src/components/FacetTabs.js
```diff
@@ -0,0 +1,153 @@
...
+      this.getTabsFromFacet(facet);
+    } else if (facetName) {
+      const searcher = this.context.searcher;
+      if (searcher && searcher.state.response) {
+        if (searcher.state.response.facets) {
+          searcher.state.response.facets.forEach((currentFacet: SearchFacet) => {
```
luigivaldez: This is really a `find()` operation on the array.... it would be clearer to call `find()` instead of `forEach()`....

src/components/FacetTabs.js
```diff
@@ -0,0 +1,153 @@
...
+      if (this.props.handleClick) {
+        tabs.push(
+          <div //eslint-disable-line
+            onClick={() => {
+              this.props.handleClick(bucket);
+            }}
+            className="facet-tab"
+          >
+            <div className="facet-tab-text">
+              {value} <span style={{ color: 'gray' }}>({count})</span>
```
luigivaldez: If you're going to have a CSS style for the tab and the text of the tab, you should use another one for the count instead of hardcoding the gray color here.

src/components/FacetTabs.js
```diff
@@ -0,0 +1,153 @@
...
+    const tabs =
+      tabData.length > 0 ? (
+        <div
+          className="facet-tab-container"
+          style={{ backgroundColor: `${this.props.backgroundColor}` }}
+        >
+          {tabData}
+        </div>
+      ) : (
+        <BusyIndicator
```
luigivaldez: What if there are just no buckets for the facet? I don't think we want to show the spinny in that case...

PR #118, opened 2019-04-25
issue-117: Adds information tooltip for Subheader360

nikhilj13:
Modifies the Subheader360 component to show an information icon that shows additional information about the components when hovered over. Here's is an example:
subheader tooltips


src/components/Subheader360.js
```diff
@@ -1,15 +1,51 @@
...
+   */
   label: string;
+  /**
+   * If true, shows an information icon next to the subheader
+   */
+  description: boolean;
+  /**
+   * The text that should be shown in the tooltip when the information icon is hovered over
+   */
+  descriptionText?: string;
```
luigivaldez: Can we rename this to "tooltip" to match the property you added to the map facets?
Also, can we avoid having a separate description Boolean property and just show the icon (and the tooltip) if the descriptionText/tooltip string is set? You can have it default to null or undefined or '' instead of "No description text provided" which we would never want to accidentally show to the user. Does this make sense?

src/components/Subheader360.js
```diff
@@ -1,15 +1,51 @@
...
-export default class Subheader360 extends React.Component<void, Subheader360Props, void> {
+type Subheader360DefaultProps = {
+  description: boolean;
+  descriptionText?: string;
+};
+
+export default class Subheader360 extends React.Component<Subheader360DefaultProps, Subheader360Props, void> {
+  static defaultProps = {
+    description: false,
+    descriptionText: 'No description text provided',
```
luigivaldez: See above.

src/components/Subheader360.js
nikhilj13: Yeah, that makes sense. I just made these changes. Let me know if you want me to change anything else :)

PR #120, opened 2019-04-25
issue-119: Update KnowledgeGraphUtils to use unique colors for unknown entities

nikhilj13:
Adds a stringToColor function that generates a unique hex color based on a stringand changes the custom color of unknown entities to be calculated using the above mentioned stringToColor function.

Knowledge graph without unique colors for entities:
kg-before
> It's difficult to differentiate between different entities in the above graph.

Knowledge graph with unique colors for entities:
kg-after
> All the different entities have unique colors which improves the readability of the above graph.


src/util/KnowledgeGraphUtils.js
```diff
@@ -327,6 +327,21 @@ export default class KnowledgeGraphUtils {
     return result;
   }
 
+  //eslint-disable-next-line
```
luigivaldez: Why is the eslint-disable needed?

src/util/KnowledgeGraphUtils.js
```diff
@@ -327,6 +327,21 @@ export default class KnowledgeGraphUtils {
...
+  static stringToColor(str) {
```
luigivaldez: Could you please add a comment in here explaining what the algorithm is doing? Thanks!

src/util/KnowledgeGraphUtils.js
nikhilj13: Because this function doesn't use the `this` keyword

src/util/KnowledgeGraphUtils.js
nikhilj13: Added comments, let me know if you need me to explain it further.

src/util/KnowledgeGraphUtils.js
nikhilj13: Lint was complaining about the function not using `this` keyword but I changed it to static as it can be useful in other cases too. Removed the lint disable comment.

PR #122, opened 2019-04-25
issue-121: Adds option to have custom location pointers and tooltip for map facet

nikhilj13:
Adds option to have custom image/Glyphicon as location pointers, also shows a tooltip with the count when you hover over a hot-spot.

For example, below is the map for customers in the United States:
mapImprovement


src/components/MapFacetContents.js
```diff
@@ -31,11 +31,20 @@ type MapFacetContentsProps = {
   size: any;
   /** The public key with which to connect to the mapbox public apis. */
   mapboxKey: string;
+  /** Custom location pointer glyphicon, default is the map-marker glyphicon */
+  glyph: string | null;
+  /** Custom location pointer image uri, if available image is shown instead of glyphicon */
+  uri: string | null;
```
luigivaldez: Can you rename this to "pointerImageUri"? And rename glyph to "pointerGlyph"?

src/components/MapFacetContents.js
```diff
@@ -31,11 +31,20 @@ type MapFacetContentsProps = {
...
+  /** Custom tooltip text to convey what the location pointers are for, for e.g. customers, transaction location etc */
+  tooltip: string;
```
luigivaldez: We should let people pass a string that is understood by the `StringUtils.fmt()` method and pass the count as the value parameter. (And then call that method later...)
This would allow the same map's tooltips to say things like "There are 4 stores here" and "There is only one store here" if the count is 4 or 1, respectively.
You could theoretically prepend the number like you're doing now if the string isn't a format string (i.e. if the format string before and after calling `StringUtild.fmt()` is the same...).
In any case, update the comment here to describe it in a bit more detail.

src/components/MapFacetContents.js
```diff
@@ -53,6 +62,9 @@ class MapFacetContents extends React.Component<MapFacetContentsDefaultProps, Map
   static defaultProps = {
     size: null,
     mapboxKey: '',
+    glyph: 'map-marker',
+    uri: null,
+    tooltip: 'occurrence(s)',
```
luigivaldez: We could update this default value to be `'{} occurrence|{} occurrences'` so it would handle the single/plural automatically.

src/components/MapFacetContents.js
```diff
@@ -210,7 +228,9 @@ class MapFacetContents extends React.Component<MapFacetContentsDefaultProps, Map
             key={`${value.longitude || 0},${value.latitude || 0}`}
             style={{ cursor: 'pointer' }}
           >
-            <Glyphicon glyph="map-marker" style={{ fontSize: '18px', color: '#2a689c' }} />
+            <OverlayTrigger overlay={<Tooltip id="tooltip-bottom">{bucket.count} {this.props.tooltip} in this region</Tooltip>}>
```
luigivaldez: It would be more readable to put the <Tooltip> tag on a new line, like this:
```
<OverlayTrigger overlay={(
  <Tooltip id="tooltip-bottom">
    {formattedTooltip}
  </Tooltip>
}}> 
  {locationPointer}
</OverlayTrigger>
```

src/components/MapFacetContents.js
```diff
@@ -198,6 +210,12 @@ class MapFacetContents extends React.Component<MapFacetContentsDefaultProps, Map
...
 
+      let locationPointer;
+      if (this.props.uri) {
+        locationPointer = <img src={this.props.uri} alt="location pointer" style={{ height: '20px', width: '20px' }} />;
+      } else if (this.props.glyph) {
+        locationPointer = <Glyphicon glyph={this.props.glyph} style={{ fontSize: '18px', color: '#2a689c' }} />;
+      }
       const points = this.props.buckets.map((bucket) => {
         const value = bucket.value; // JSON.parse(bucket.value);
         // Keep track of the boundaries of the coordinates
```
luigivaldez: Here is where you might add the tooltip calculation, like:
```
let formattedTooltip;
formattedTooltip = StringUtils.fmt(this.props.tooltip, bucket.count);
if (formattedTooltip === this.props.tooltip) {
  formattedTooltip = `${bucket.count} ${this.props.tooltip}`;    
}
```

luigivaldez (changes requested):

src/components/MapFacetContents.js
nikhilj13: Done

src/components/MapFacetContents.js
nikhilj13: That makes sense, I wasn't aware of that function in StringUtils. I changed the tooltips to use StringUtils.fmt() and updated the comment as well.

src/components/MapFacetContents.js
nikhilj13: Done

src/components/MapFacetContents.js
nikhilj13: Done

src/components/MapFacetContents.js
```diff
@@ -31,19 +31,25 @@ type MapFacetContentsProps = {
...
   /** Custom location pointer image uri, if available image is shown instead of glyphicon */
-  uri: string | null;
-  /** Custom tooltip text to convey what the location pointers are for, for e.g. customers, transaction location etc */
+  pointerImageUri: string | null;
+  /**
+   * Custom tooltip text to convey what the location pointers are for, for e.g. customers, transactions etc
+   * It should specify the singular and plural forms to be used
+   * For example a tooltip for cats would be '{} cat|{} cats'
+   * If the string is not in the format specified above,
+   * it would be used as it is concatenated with the count.
```
luigivaldez: I might just change "concatenated with" to be "appended to" so it's unambiguous that the order will be first the count and then the tooltip string.

luigivaldez (changes requested):
Last nitpicky thing. :)

src/components/MapFacetContents.js
nikhilj13: Changed!

luigivaldez (approved):

PR #125, opened 2019-05-02
Update date format

shetty-shruti:
Updating DateUtils and DocumentEntityList where when dates are converted to desired format with midnight date, current format converts it to a previous day on UI. Issue raised against JIRA ticket NYL-352


src/components/DocumentEntityList.js
```diff
@@ -38,7 +38,8 @@ export default class DocumentEntityList extends React.Component<void, DocumentEn
       dataPairs = dataPairs.filter((dataPair: DataPairInfo) => {
         return dataPair.type !== FieldNames.DATE;
       });
-      const dateString = DateUtils.formatDateString(date, DateFormat.LONG_DATE);
+      // Updating dateFormat for
```
luigivaldez: Was there supposed to be more to this comment?

src/util/DateUtils.js
```diff
@@ -80,6 +80,36 @@ export default class DateUtils {
...
+
+   /**
+   * Given a string, date format and locale, return a formatted date in the specified locale using
+   * the format specified.
+   * @param dateString date as string
+   * @param format given format for coversion
+   * @param locale given locale
+   * @return a formatted date in the specified locale using
+   * the format specified.
+   * @author Shruti Shetty
```
luigivaldez: We don't generally add `@author` tags in the JSDoc comments

src/util/DateUtils.js
```diff
@@ -80,6 +80,36 @@ export default class DateUtils {
     return '';
   }
 
+  /**
+   * Given a string, return a Date object created by parsing it.
```
luigivaldez: The comment mentions nothing about UTC... Does the original string represent a UTC time? If so, maybe we can call the function `utcStringToDate()`? And/or name the parameter `utcDateString`?

Can you maybe add an example of a string you might pass to it? Maybe in a @param tag?

src/util/DateUtils.js
```diff
@@ -80,6 +80,36 @@ export default class DateUtils {
...
+   * Given a string, date format and locale, return a formatted date in the specified locale using
+   * the format specified.
+   * @param dateString date as string
+   * @param format given format for coversion
+   * @param locale given locale
+   * @return a formatted date in the specified locale using
+   * the format specified.
+   * @author Shruti Shetty
+   */
+
```
luigivaldez: I'd remove this extra line so the comment is right next to the method...

src/util/DateUtils.js
```diff
@@ -80,6 +80,36 @@ export default class DateUtils {
...
+  static formatDateStringToUTC(
```
luigivaldez: Can you describe what the reason for having this method is as opposed to letting users call
`DateUtils.formatDate(DateUtils.stringtoUTCDate(dateString), format, locale);`
I'm wondering how commonly used this would be... if it's just for the DocumentEntityList component, maybe it doesn't warrant being a separate method?

klk1010:
Stale

PR #129, opened 2019-05-28
Bug/plat 43324

klk1010:
Addresses part of [PLAT-43324](https://jira.attivio.com/browse/PLAT-43324).

Updates FetchUtils.fetch to properly redirect to login on auth failure. Previously, when an html login response was returned, the call would log a console error `REST call to ${uri} failed to return JSON.` Now there as an additional check to examine the html to see if it's html for login. If it is, the page is reloaded, triggering the login page. If it isn't, the fetch falls back to the existing error: `REST call to ${uri} failed to return JSON.`

This was causing in issue in Connector Admin because it uses`<Searcher />` and `<Searcher />` uses the fetch utils defined here.


src/util/FetchUtils.js
```diff
@@ -46,6 +46,8 @@ export default class FetchUtils {
                 // that it's a redirect to a login page. If so, then we then reload the whole
                 // page to let the user log in again.
                 window.location.reload();
+              } else if (text.startsWith('<html') && text.includes('j_security_check')) {
```
luigivaldez: Isn't it possible for there to be a doctype at the start of the text? Would it make sense to check for the presence of '<html' instead of checking for it at the beginning?

klk1010: Perhaps, I copied that conditional from Search Analytics. I could probably just do the `text.includes('j_security_check')`.

PR #131, opened 2019-06-14
bug/PLAT-43823: Distinguish between toggled on disabled and toggled off

klk1010:
Addresses remainder of [PLAT-43823](https://jira.attivio.com/browse/PLAT-43823)

`<ToggleSwitch />` now distinguishes between disabled off and disabled on.

Screen Shot 2019-06-14 at 1 41 48 PM


src/components/ToggleSwitch.js
```diff
@@ -49,125 +49,168 @@ export default class ToggleSwitch extends React.Component<ToggleSwitchDefaultPro
...
-          />
-        </div>
-      );
-    }
+          className={slider.className}
+          style={slider.style}
+        />
+        <div
+          className={button.className}
+          onClick={this.onClick}
```
luigivaldez: One thing I noticed when poking around in the UI was that for the no-label variation, you have to click the circle to toggle the switch. I think it would be useful to be able to click anywhere on the component to do this... which could be done by just moving the onClick to the outer `<div>`, I think. Thoughts?

klk1010: makes sense, I'll change it

PR #132, opened 2019-06-18
bug/PLAT-43828 changes made to display only selected connectors onload & on connector creation instead of all connectors

VeenaIyer:
bug/PLAT-43828 On ConnectorAdmin page load, if Stop selected is clicked from Selected Items menu, all Connectors are prompted
Reload / Open connector admin in a new tab. Click on "Selected Items" gear icon, and select any action. The Confirmation box shows all the connectors, even though by default, only the first connector is selected.

Changes made to ensure on page load & on creation on new connector if user does not select any connector, the first connector is only selected and not all the connectors. 

image

### DIFF
--- a/src/components/Table.js
+++ b/src/components/Table.js
@@ -283,9 +283,12 @@ export default class Table extends React.Component<TableDefaultProps, TableProps
       })
       : [];
       this.setState({ sortedRows }, () => {
-        const activeRow = sortedRows[this.state.activeRowIndex];
+        const activeRow = this.state.activeRowIndex !== null && newProps.noEmptySelection
+        ? sortedRows[this.state.activeRowIndex] : null;
+        const selectedRows = this.state.activeRowIndex !== null && newProps.noEmptySelection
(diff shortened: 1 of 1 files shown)


src/components/Table.js
```diff
@@ -283,9 +283,12 @@ export default class Table extends React.Component<TableDefaultProps, TableProps
       })
       : [];
       this.setState({ sortedRows }, () => {
-        const activeRow = sortedRows[this.state.activeRowIndex];
+        const activeRow = this.state.activeRowIndex !== null && newProps.noEmptySelection
```
luigivaldez: If the `noEmptySelection` flag is true, won't we end up with no selection if the previous activeRowIndex was null because there were no items, and now we have items? For example, if I filter out everything and then I delete the filter, should the first item not be selected/active?

benmvachon: This is a good point. The same logic is used on [line 268](https://github.com/attivio/suit/pull/132/files#diff-7d12218e29d596148ef20ac652d3a3b9L268).

PR #130, opened 2019-06-03
task/PLAT-43404 : Adding data-test attribute to uniquely identify elements for Selenium

shah-tejas:
PLAT-43404 : Adding data-test attribute to uniquely identify UI elements for Selenium.

Added data-test attribute to certain UI elements from SUIT, that are used in ConnectorAdmin.

PR #133, opened 2019-06-20
task/PLAT-43404 : Adding support for components to add data-test attribute to uniquely identify elements for Selenium

shah-tejas:
[PLAT-43404](https://jira.attivio.com/browse/PLAT-43404) : Add id attributes to make Selenium testing easier

This is a continuation of https://github.com/attivio/suit/pull/130

Added support for data-test attribute for components in SUIT, that are used in ConnectorAdmin.


src/components/Menu.js
```diff
@@ -105,6 +105,12 @@ type MenuProps = {
   width: number | null;
   /** Any CSS style you need to apply to the menu. Optional. */
   style: any;
+  /**
+   * Property that contains the value for data-test attribute
+   * added to elements to be uniquely identified by testing tools
+   * like Selenium.
+   */
+  dataTestValue?: string | null;
```
luigivaldez: I'm wondering if it makes sense to call this `dataTest` instead of `dataTestValue` since the eventual HTML attribute is called `data-test`? This would be in keeping with other pass-through things like `id` or `style` where the React component property has the same name as the HTML attribute...

src/components/Menu.js
```diff
@@ -280,7 +289,7 @@ export default class Menu extends React.Component<MenuDefaultProps, MenuProps, v
       ) : null;
 
     const buttonLabel = (
-      <span>
+      <span data-test={dataTestValue}>
```
luigivaldez: Should the data-test attribute be on this `<span>` or on the < Dropdown.Toggle> component in the render() method? That would also include the little down-facing caret and the border of the button that triggers the menu. This way, the data-test info will only be available if the user clicks on the span itself.

src/components/TabPanel.js
```diff
@@ -28,17 +28,25 @@ type TabPanelProps = {
    * class assigned, for use in nested search results.
    */
   nested: boolean;
+  /**
+   * Property that contains the prefix for data-test attribute
+   * added to elements to be uniquely identified by testing tools
+   * like Selenium.
```
luigivaldez: In the comment, you should say what will be appended to the prefix to create the final value of the data-test attribute. For example, the `id` field of the TabInfo object for each tab in this case. This way someone will know that, given the prefix "inspectorTab-" and the tabs "a," "b," and "c," the actual data-test attributes will be "inspectorTab-a," "inspectorTab-b," and "inspectorTab-c."

src/components/ToggleSwitch.js
```diff
@@ -23,6 +23,12 @@ type ToggleSwitchProps = {
    * as well.
    */
   className?: string;
+  /**
+   * Property that contains the value for data-test attribute
+   * added to elements to be uniquely identified by testing tools
+   * like Selenium.
+   */
+  dataTestValue?: string | null;
```
luigivaldez: Here, too... I think `dataTest` is better...

src/components/Menu.js
shah-tejas: Yes, makes sense to have `data-test` on the `<Dropdown.Toggle>` component. Will make the changes.

shah-tejas:
Addressed code review comments.

PR #136, opened 2019-07-12
bug/PLAT-41437 changes to ensure no action is performed on double click of selected node

VeenaIyer:
PLAT-41437 Clicking on main document in knowledge graph of 360 view gives an exception

Changes to ensure double click of selected node (patient 0) does not lead to any action.

### DIFF
--- a/src/components/KnowledgeGraphPanel.js
+++ b/src/components/KnowledgeGraphPanel.js
@@ -129,10 +129,18 @@ export default class KnowledgeGraphPanel extends React.Component<KnowledgeGraphP
           const node = this.state.nodes.find((n) => {
             return n.id === nodeId;
           });
-          if (node && node.docId) {
-            this.props.navigateToDoc(node.docId);
-          } else if (node) {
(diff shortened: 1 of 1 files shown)


src/components/KnowledgeGraphPanel.js
```diff
@@ -129,10 +129,18 @@ export default class KnowledgeGraphPanel extends React.Component<KnowledgeGraphP
...
-          if (node && node.docId) {
-            this.props.navigateToDoc(node.docId);
-          } else if (node) {
-            this.props.navigateToEntity(node.group, node.label);
+          /**
+           * we do not want any action to be performed
+           * on the double click of the selected node
+           * i.e. the zeroth node.
+           * */
+          if (nodeId !== 0) {
```
luigivaldez: We could even put this check before line 129. There's no reason to look through the nodes to get the node object if we're never going to use it...

VeenaIyer: True, I thought it is safer to still have to find the selected node but yes we don't want any action to be performed.

PR #138, opened 2019-07-12
issue-137: Adds an option to add comments on a Search Result

nikhilj13:
Adds the functionality to add/view comments on a search result.

Here's an example where I'm adding and deleting a comment:
comment

The comments are stored in the index and the users would be able to enable/disable comments and specify the table name for comment documents in the Index. These options would be configurable from the SearchUI's configuration.properties file. Below is an example:
```
// These properties configure the tags/options shown for each of the list search results
  SearchResultTags: {
	  // Whether to show an option to add/view comments for a search result. False by default.
	  comments: false,
	  // Specify the table field for the comment documents in the index. 'comments' by default.
	  commentsTable: 'comments'
  },
```


src/components/Comments.js
```diff
@@ -0,0 +1,253 @@
+import React from 'react';
```
klk1010: This is missing the flow annotation `// @flow`

klk1010: Add the flow annotation.
```suggestion
// @flow
import React from 'react';
```

src/components/Comments.js
```diff
@@ -0,0 +1,253 @@
...
+    (this: any).showCommentModal = this.showCommentModal.bind(this);
+    (this: any).hideCommentModal = this.hideCommentModal.bind(this);
+    (this: any).captureComment = this.captureComment.bind(this);
+    (this: any).saveComment = this.saveComment.bind(this);
+    (this: any).clearComment = this.clearComment.bind(this);
+    (this: any).getComments = this.getComments.bind(this);
+    (this: any).deleteComment = this.deleteComment.bind(this);
+  }
+
+  componentWillMount() {
```
klk1010: Use `componentDidMount` instead of `componentWillMount`.
https://reactjs.org/docs/react-component.html#mounting

src/components/Comments.js
```diff
@@ -0,0 +1,253 @@
...
+  }
+
+  render() {
+    const noCommentLabel = (
+      <div style={{ width: '100%', textAlign: 'center', padding: '1em', color: 'gray' }}>
+        <i> No Comments Available </i>
+      </div>
+    );
+
+    const commentModal = (
```
klk1010: Pull this out of render into its own function, `renderCommentModal`

src/components/Comments.js
```diff
@@ -0,0 +1,253 @@
...
+  }
+
+  clearComment() {
+    this.setState({
+      comment: '',
+    });
+  }
+
+  render() {
+    const noCommentLabel = (
```
klk1010: Pull this out of render into its own function `renderCommentLabel`

src/components/Comments.js
```diff
@@ -0,0 +1,253 @@
...
+        </Modal.Body>
+        <Modal.Footer>
+          <Button onClick={this.hideCommentModal}>
+            Close
+          </Button>
+        </Modal.Footer>
+      </Modal>
+    );
+
+    const commentCount = this.state.commentList && this.state.commentList.length > 0 ? this.state.commentList.length : '';
```
klk1010: Default to `0` instead of `''`.

src/components/Comments.js
```diff
@@ -0,0 +1,253 @@
...
+    const commentLabel = commentCount > 0 ? `Comment (${commentCount})` : 'Comment';
+    const commentLink = (
```
klk1010: Break out to separate render function `renderCommentLink`.

src/components/SearchResultTags.js
```diff
@@ -39,6 +40,10 @@ type SearchResultTagsProps = {
    * Defaults to the value in the configuration.
    */
   baseUri?: string,
+  /** Whether to show a Comment option. Defaults to false. */
+  comments: boolean;
```
klk1010: Indicate that this prop is optional in the flow annotation.
```suggestion
  comments?: boolean;
```

src/components/SearchResultTags.js
```diff
@@ -39,6 +40,10 @@ type SearchResultTagsProps = {
...
+  /** Table field for the comment documents */
+  commentsTable: string | null;
```
klk1010: Indicate commentsTable is an optional prop in the flow annotation. Note, that passing commentsTable as null is not the same as not providing commentsTable prop at all.
```suggestion
  commentsTable: string | null;
```

src/components/SearchResultTags.js
```diff
@@ -317,10 +324,15 @@ class SearchResultTags extends React.Component<SearchResultTagsDefaultProps, Sea
       </span>
     ) : '';
 
+    const comments = this.props.comments ? (
```
klk1010: This could just be: 
```
    const comments = this.props.comments && (
      <Comments docId={this.props.docId} commentsTable={this.props.commentsTable} />
    );
```

src/components/Comments.js
```diff
@@ -0,0 +1,272 @@
...
+        <i> No Comments Available </i>
+      </div>
+    );
+    const comments = commentList.length > 0 ? commentList : noCommentLabel;
+    return comments;
+  }
+
+  renderCommentModal() {
+    const { showCommentModal, comment } = this.state;
+    const commentModal = (
```
klk1010: There's no reason to save this as a variable. Instead, just return this directly.
```
return (
  <Modal
    ...
  </Modal>
);
```

src/components/Comments.js
```diff
@@ -0,0 +1,272 @@
...
+      });
+    }
+  }
+
+  deleteComment(id: string) {
+    const jsonDoc = `{ "id" : "${id}", "mode" : "DELETE" }`;
+    this.context.searcher.search.addOrDeleteDocument(JSON.parse(jsonDoc), this.getComments);
+  }
+
+  populateComments() {
```
klk1010: Instead of storing JSX as state, it would be better to just render this and set a flag to determine if it should be shown or not.

src/components/Comments.js
```diff
@@ -0,0 +1,272 @@
...
+    const { searcher } = this.context;
+    const username = AuthUtils.getUserName(AuthUtils.getSavedUser());
+    const d = new Date();
+    const timestamp = `Posted on ${d.toLocaleDateString()} at ${d.toLocaleTimeString()}`;
+    const loggedDateTime = d.toISOString();
+    const id = username.concat(loggedDateTime);
+    // escape newline, carriage return and tab characters in the comment string
+    const formattedComment = comment.replace(/\n/g, '\\n').replace(/\r/g, '\\r').replace(/\t/g, '\\t');
+
+    const jsonDoc = `{ "fields" : { "comment_s" : [ "${formattedComment}" ], `
```
klk1010: You should be able to do this as an object - much easier to read, and use `JSON.stringify`.
```
const body = { 
  fields: {
    comment_s: [formattedComment],
    docId_s: [docId],
    username_s: [username],
    date: [loggedDateTime],
    timestamp_s: [timestamp],
    table: ["comments"]
  },
  id,
};
```

src/components/Comments.js
```diff
@@ -0,0 +1,272 @@
...
+
+  clearComment() {
+    this.setState({
+      comment: '',
+    });
+  }
+
+  renderComments() {
+    const { commentList } = this.state;
+    const noCommentLabel = (
```
klk1010: This is a bit more readable, but getting pretty nit picky: Just makes it a bit clearer that you want to show comments, but you're providing a fallback.
```
const showComments = commentList && commentList.length > 0;
if (showComments) {
  return commentList;
}
return (
  <div style={{ width: '100%', textAlign: 'center', padding: '1em', color: 'gray' }}>
    <i> No Comments Available </i>
  </div>
);
```

src/components/Comments.js
```diff
@@ -0,0 +1,272 @@
...
+    const jsonDoc = `{ "id" : "${id}", "mode" : "DELETE" }`;
+    this.context.searcher.search.addOrDeleteDocument(JSON.parse(jsonDoc), this.getComments);
+  }
+
+  populateComments() {
+    const { response } = this.state;
+    const loggedInUser = AuthUtils.getUserName(AuthUtils.getSavedUser());
+    if (response) {
+      const commentList = [];
+      response.documents.forEach((doc) => {
```
klk1010: The response formatting should be done in a generic location. We follow a pattern of saving requests in the `api` directory - as you did - and also creating a js class created from the result. `api/SimplePrincipal` is an example of this. If you do that on line 71 so that instead of storing `response` in state directly, you're storing `commentList`, which would be: `Array<Comment>`, then here you could just map the result and `Comment` would be it's own `.js` file and class that could potentially be used elsewhere.

src/components/Comments.js
nikhilj13: What if I don't store it as state but instead call this function from render?

src/api/Search.js
```diff
@@ -212,4 +212,102 @@ export default class Search {
         });
     });
   }
+
+  addOrDeleteDocument(jsonRequest: JSON, callback: () => void): Promise<any> {
```
klk1010: Should this be `addOrDeleteComment` or is this endpoint more generic than that?

nikhilj13: It is more generic.. you can add or delete any document by specifying the table field.

src/components/SearchResultTags.js
klk1010: Still needs to marked as optional here.
```suggestion
  comments?: boolean;
```

src/components/Comments.js
```diff
@@ -0,0 +1,268 @@
...
+    const d = new Date();
+    const loggedDateTime = d.toISOString();
+    const id = username.concat(loggedDateTime);
+    const body = {
+      fields: {
+        comment_s: [comment],
+        docId_s: [docId],
+        username_s: [username],
+        date: [loggedDateTime],
+        table: ['comments'],
```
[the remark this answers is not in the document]
nikhilj13: Wouldn't the users need to create such a zone first? I think it makes sense to have a separate zone considering we don't want these docs to show as a part of the search results. 

If we configure such a zone by default and use it for saved searches, comments and briefcases (upcoming PR that would add the ability to store search results in baskets/buckets/briefcases), it would make things more consistent. Thoughts?

src/components/Comments.js
```diff
@@ -0,0 +1,268 @@
...
+    this.context.searcher.search.addOrDeleteDocument(JSON.parse(jsonDoc), this.getComments);
+  }
+
+  createFormattedCommentList() {
+    const { commentList } = this.state;
+    const loggedInUser = AuthUtils.getUserName(AuthUtils.getSavedUser());
+    const formattedCommentList = [];
+    if (commentList && commentList.length > 0) {
+      commentList.forEach((comment) => {
+        const isCommentByLoggedInUser = (comment.username === loggedInUser);
```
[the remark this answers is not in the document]
nikhilj13: I believe it's the full name.. `AuthUtils.getSavedUser()` gives the userId and `AuthUtils.getUserName(AuthUtils.getSavedUser())` gives the full name.

src/api/Comment.js
```diff
@@ -0,0 +1,39 @@
...
+import SearchDocument from './SearchDocument';
+import DateUtils from '../util/DateUtils';
+
+/**
+ * A single comment for a SearchDocument.
+ */
+export default class Comment {
+  static createTimestamp(date) {
+    const d = DateUtils.stringToDate(date);
+    return `Posted on ${d.toDateString()} at ${d.toLocaleTimeString()}`;
```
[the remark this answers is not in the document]
nikhilj13: The timestamp does specify the time zone, `toLocaleTimeString()` handles that.

src/components/Comments.js
```diff
@@ -0,0 +1,268 @@
...
+          />
+        );
+        formattedCommentList.push(
+          <div key={comment.id}>
+            <div
+              style={{ color: '#484848', backgroundColor: '#F5F5F5', borderRadius: '5px' }}
+              title={comment.timestamp}
+            >
+              <div style={{ padding: '0.5em' }}>
+                <b> <i> {comment.username} </i> </b>
```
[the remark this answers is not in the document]
nikhilj13: I tried that first but the username just blends into the comment. But I guess I can make the username less prominent.

src/components/Comments.js
```diff
@@ -0,0 +1,268 @@
...
+    const jsonDoc = `{ "id" : "${id}", "mode" : "DELETE" }`;
+    this.context.searcher.search.addOrDeleteDocument(JSON.parse(jsonDoc), this.getComments);
+  }
+
+  createFormattedCommentList() {
+    const { commentList } = this.state;
+    const loggedInUser = AuthUtils.getUserName(AuthUtils.getSavedUser());
+    const formattedCommentList = [];
+    if (commentList && commentList.length > 0) {
+      commentList.forEach((comment) => {
```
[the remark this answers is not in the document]
nikhilj13: Yeah, I an use map() instead.. it's slightly faster than forEach().

src/components/Comments.js
```diff
@@ -0,0 +1,268 @@
...
+      <Modal
+        show={showCommentModal}
+        onHide={this.hideCommentModal}
+        backdrop="static"
+      >
+        <Modal.Header closeButton>
+          <Modal.Title>Add or View Comments</Modal.Title>
+        </Modal.Header>
+        <Modal.Body>
+          <Row>
```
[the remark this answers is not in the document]
nikhilj13: I think if there are no comments then add comments box should be on top.. otherwise the comments.

tests/api/Comment-test.js
```diff
@@ -0,0 +1,134 @@
+import expect from 'expect';
+import Comment from '../../src/api/Comment';
+import SearchDocument from '../../src/api/SearchDocument';
+
+// We don't need to lint the JSON data...
+/* eslint-disable quote-props,comma-dangle,quotes,max-len */
+const testDoc1 = {
```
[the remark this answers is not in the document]
nikhilj13: I agree, I'll change the test document.

src/components/Comments.js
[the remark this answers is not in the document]
nikhilj13: Yeah I think it makes sense. But then shouldn't there be a notification or a way to inform the user whose comment got deleted and why it got deleted.

src/api/Search.js
```diff
@@ -212,4 +212,102 @@ export default class Search {
...
+
+              const body = JSON.stringify(jsonRequest);
+              const params = {
+                method: 'POST',
+                headers,
+                body,
+                credentials: 'include',
+              };
+
+              const updateFetchRequest = new Request(updateUri, params);
```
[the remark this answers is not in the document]
nikhilj13: I don't think the FetchUtils' fetch function does what I'm trying to do here...

PR #139, opened 2019-07-12
task/PLAT-44055 : Create signal when user selects an autocomplete entry

shah-tejas:
[PLAT-44055](https://jira.attivio.com/browse/PLAT-44055): Add signals to autocomplete

Add a signal using `/rest/signals/add` endpoint whenever user clicks/selects an autocomplete entry.
This will be useful in searchui to create signals of type `autocomplete`. The `autoCompleteSignalType` prop will need to be passed to the `SearchBar` component from searchui to ensure the signal is created.


src/components/AutoCompleteInput.js
```diff
@@ -141,12 +150,31 @@ export default class AutoCompleteInput extends React.Component<AutoCompleteInput
     }
   }
 
+  updateValueAndAddSignal(newQuery: string, doSearch: boolean = false) {
+    const docOrdinal = this.state.suggestions.findIndex((suggestion) => {
```
klk1010: Code styling comment, destructure props and state at top of function.
```
const { suggestions, queryTimestamp, userInput } = this.state;
const { updateValue } = this.props;
```
```

File: src/components/SearchBar.js [hunk 2]
```diff
@@ -54,6 +58,10 @@ type SearchBarProps = {
   buttonLabel: string;
   /** If set, this is the route to navigate to upon executing a search. By default, no navigation will occur when searching. */
   route: string | null;
+  /**
+   * If set, a new signal of this type would be added when an autocomplete item is selected.
+   */
+  autoCompleteSignalType: string | null;
```
klk1010: Indicate `autoCompleteSignalType` is optional with the flow annotation.
```suggestion
  autoCompleteSignalType?: string | null;
```

src/components/SearchBar.js
```diff
@@ -175,9 +186,37 @@ class SearchBar extends React.Component<SearchBarDefaultProps, SearchBarProps, S
     }
   }
 
-  updateQuery(newQuery: string, doSearch: boolean = false) {
+  addSignal(query: string, signalData: any) {
```
klk1010: Instead of `signalData: any`, should this be `signalData: SignalData`?

shah-tejas:
Code comments addressed.

PR #140, opened 2019-07-16
bug/PLAT-41442 unifying behaviour of search on the click of Enter & Go and having the facet filter reset on new query

VeenaIyer:
PLAT-41442 Facet Filters Not removed when clicked on 'Go' after query is entered

On the click of Go/Enter, check is performed to see if search for that particular query has been carried out.
If yes, hard reset is avoided and only search is performed.
If no, means it is a new search and thus old filters are removed, hard reset and search is performed.

In each case search i.e. the searcher.doSearch is called as it can happen that some ingestion is being done and a click of Go/Enter for the same term can fetch in more results.

Can add more comments if more clarity is needed in code.

### DIFF
--- a/src/components/SearchBar.js
+++ b/src/components/SearchBar.js
@@ -219,8 +219,12 @@ class SearchBar extends React.Component<SearchBarDefaultProps, SearchBarProps, S
     }
     if (searcher) {
       if (doSearch) {
-        searcher.setQueryAndSearch(newQuery);
-        this.route();
+        if (!searcher.state.haveSearched) {
+          searcher.setQueryAndSearch(newQuery);
+          this.route();
+        } else {
+          searcher.doSearch();
+        }
       } else {
         searcher.updateQuery(newQuery);
       }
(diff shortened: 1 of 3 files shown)


src/components/SearchBar.js
```diff
@@ -219,8 +219,12 @@ class SearchBar extends React.Component<SearchBarDefaultProps, SearchBarProps, S
     }
     if (searcher) {
       if (doSearch) {
-        searcher.setQueryAndSearch(newQuery);
-        this.route();
+        if (!searcher.state.haveSearched) {
```
luigivaldez: Please add a comment here explaining what happens and in which cases...

VeenaIyer:
> Did it turn out to be the case that if the query string changes, the `haveSearched` flag will be false? That's a good observation and it means you don't have to do anything special to compare the old and new query strings or whatever, right?
> I just had one comment (about a comment)...

> Yup so whenever a user starts typing or changes anything the haveSearched flag is updated to false. Thus no comparison needed.
> Will add the comment.

src/components/SearchBar.js
VeenaIyer: Done!

PR #141, opened 2019-07-17
Bug/plat 44063 a

klk1010:
Addresses [PLAT-44063](https://jira.attivio.com/browse/PLAT-44063)

Fixes issue in connectoradmin where running a connector would cause the table selection to reset, causing it to be inconsistent with the connector displayed in the inspector details.

### DIFF
--- a/src/components/Table.js
+++ b/src/components/Table.js
@@ -241,48 +241,97 @@ export default class Table extends React.Component<TableDefaultProps, TableProps
   }
 
   componentWillReceiveProps(newProps: TableProps) {
-    const { rows, noEmptySelection } = newProps;
+    const { rows, noEmptySelection, rowComparator, onSelect } = newProps;
     const { rows: prevRows } = this.props;
+    const { selectedIndices: prevSelectedIndices, activeRowIndex } = this.state;
 
     // If a rowComparator is specified, use it, otherwise, do a deep row comparison and reset selection if anything changed.
     const shouldUpdateRows = !isEqual(rows, prevRows);
-    const shouldResetSelection = rows.length !== prevRows.length
-      || (newProps.rowComparator && !isEqualWith(rows, prevRows, newProps.rowComparator))
-      || (!newProps.rowComparator && shouldUpdateRows);
+    let shouldResetSelection = (rowComparator && !isEqualWith(rows, prevRows, rowComparator))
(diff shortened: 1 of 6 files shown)


src/components/Table.js
```diff
@@ -241,46 +241,72 @@ export default class Table extends React.Component<TableDefaultProps, TableProps
...
+      || (!rowComparator && shouldUpdateRows);
 
     if (shouldResetSelection || shouldUpdateRows) {
+      const updatedSelectedRows = [];
+
       const sortedRows = rows && rows.length > 0
         ? rows.map((row, tableRowIndex) => {
-          return { ...row, tableRowIndex };
+          const updatedRow = { ...row, tableRowIndex };
+          if (prevSelectedIndices.has(tableRowIndex)) {
```
luigivaldez: In this loop, you're adding any previously selected rows to the updated selection if they're still in the list, right? Maybe call this out explicitly in a comment?

src/components/Table.js
```diff
@@ -241,46 +241,72 @@ export default class Table extends React.Component<TableDefaultProps, TableProps
...
+      // Indicates one of the previously selected rows can no longer be found.
+      if (numPreviousSelections !== numUpdatedSelections) {
+        shouldResetSelection = true;
+      }
+
+      const resetIndices = sortedRows.length > 0 && noEmptySelection
         ? new Set([0])
         : new Set([]);
-      const anchorRowIndex = sortedRows.length > 0 && noEmptySelection
+      const resetIndex = sortedRows.length > 0 && noEmptySelection
```
luigivaldez: If a row that had been selected is no longer in the table, but the previous anchor row is still there, won't this change the anchor row? I could be missing something—there are a lot of moving parts... :)

klk1010: If the previously selected row is no longer in the table, we _want_ to reset the anchor row. The anchor row is only meaningful in relation to the selected row.

luigivaldez: Oh, I think I was thinking that the anchor row was the one shown in the inspector... OK.

klk1010: Ah, gotchya. The `activeRowIndex` indicates the row that's displayed in the inspector. The `anchorRowIndex` is an internal tracking mechanism for multiselect keyboard actions to determine direction and boundaries.

PR #146, opened 2019-07-22
Chore/plat 0 react 16.8.6

klk1010:
# Updates suit to react 16.8.6.

Addresses [PLAT-44155](https://jira.attivio.com/browse/PLAT-44155)

* Updates `react-dom` to 16.8.6 - Must always have the same version as react.
* Updates `react-dates`. 
  - The old version relied on deprecated `react-addons-shallow-compare` as a peer dependency. That library no longer works in React 16 because it references react internals in react/lib, which no longer exists.
  - Required addition of imports to maintain current styling.
* Fixed linting errors (unrelated to these changes).
* Updated ObjectUtils.deepEquals to just use `isEqual` from `lodash`.

To update searchui: https://github.com/attivio/searchui/pull/86


klk1010:
Moving this up to 1.0.0 is not a bad idea. I'll do that before merging.

PR #149, opened 2019-08-01
task/PLAT-44214 : Add signal when a facet filter is added/removed

shah-tejas:
[PLAT-44214](https://jira.attivio.com/browse/PLAT-44214) : Add signals to facet filters

The `addFacetFilter()` and `removeFacetFilter()` functions in `Searcher.js` are called every time a facet is added / removed in the search. Call to `addSignal()` is added in these functions before search is performed with the updated filters.
Signal is created with weight 1 when adding a Filter and with weight 0 when removing a Filter.

### DIFF
--- a/src/components/Searcher.js
+++ b/src/components/Searcher.js
@@ -11,8 +11,12 @@ import SimpleQueryRequest from '../api/SimpleQueryRequest';
 import QueryResponse from '../api/QueryResponse';
 import FacetFilter from '../api/FacetFilter';
 import FieldNames from '../api/FieldNames';
+import SignalData from '../api/SignalData';
+import SearchDocument from '../api/SearchDocument';
+import Signals from '../api/Signals';
 
 import ObjectUtils from '../util/ObjectUtils';
+import AuthUtils from '../util/AuthUtils';
 
 import Configurable from '../components/Configurable';
 import Configuration from '../components/Configuration';
@@ -204,6 +208,7 @@ type SearcherState = {
   resultsPerPage: number;
   resultsOffset: number;
   debug: boolean;
+  queryTimestamp: number;
 };
 
 /**
@@ -391,6 +396,7 @@ class Searcher extends React.Component<SearcherDefaultProps, SearcherProps, Sear
       resultsPerPage: parseInt(this.props.resultsPerPage, 10),
       resultsOffset: 0,
       debug: this.props.debug,
+      queryTimestamp: 0,
     };
   }
 
@@ -503,11 +509,13 @@ class Searcher extends React.Component<SearcherDefaultProps, SearcherProps, Sear
     delete currentState.error;
     delete currentState.response;
     delete currentState.haveSearched;
+    delete currentState.queryTimestamp;
 
     const newState = Object.assign({}, compareWith);
     delete newState.error;
     delete newState.response;
     delete newState.haveSearched;
+    delete newState.queryTimestamp;
 
     return !ObjectUtils.deepEquals(currentState, newState);
   }
@@ -680,6 +688,7 @@ class Searcher extends React.Component<SearcherDefaultProps, SearcherProps, Sear
       relevancyModels,
       debug,
       haveSearched: this.state.haveSearched, // Make sure we don't change this
+      queryTimestamp: 0,
     };
 
     return result;
@@ -738,6 +747,7 @@ class Searcher extends React.Component<SearcherDefaultProps, SearcherProps, Sear
         response,
         error: undefined,
         haveSearched: true,
+        queryTimestamp: Date.now(),
       });
     } else if (error) {
       // Failed!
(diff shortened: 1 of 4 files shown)


src/components/Searcher.js
```diff
@@ -970,6 +990,44 @@ class Searcher extends React.Component<SearcherDefaultProps, SearcherProps, Sear
...
+    }
+    const facet = facets.find((searchFacet) => {
+      return searchFacet.label === facetFilter.facetName;
+    });
+    if (!facet) {
+      return;
+    }
+    const signal = new SignalData();
+    signal.docId = facetFilter.filter;
+    signal.docOrdinal = facet.buckets.findIndex((bucket) => {
```
luigivaldez: If the query returned documents, I'm wondering if we should try to get most of this information out of the signal returned for the first document. This will provide much more accurate information and allow tools to better associate these facet filter signals with others. The query, queryTimestamp, relevancyModelName, etc. fields, for example, will be different than what you can get from the searcher...

What do you think about this idea?

shah-tejas: Would be nice to use that. But which fields would be relevant from that signal? Only ones I can think of are locale, principal.

luigivaldez: Actually, I think a lot of them would be...
* principal
* query
* queryTimestamp
* relevancyModelName
* relevancyModelNames

And maybe locale, too...

For example, the query field could be different than the string that the searcher knows about because the back-end code can transform it and using the one that will end up in click signals will make the click signals and facet signals easier to match up later. The same for the queryTimestamp.... knowing the queryTimestamp from the original query will also help match things up.

We could save these fields instead of just the query time you're currently saving (since that won't match the one in the signals anyway). If there are no documents returned, then there won't be any applicable facet filters, so we should always have the data available when submitting a facet signal, I believe.

shah-tejas: Oh yeah, that makes sense. For some reason I thought the query (and therefore queryTimestamp) would change when we apply filter. But my bad.
Looks like most of the properties from querySignal can be easily used here.

src/components/Searcher.js
```diff
@@ -204,6 +207,7 @@ type SearcherState = {
   resultsPerPage: number;
   resultsOffset: number;
   debug: boolean;
+  queryTimestamp: number;
```
luigivaldez: I don't think we still need this if you're using the signal from the search results.

src/components/Searcher.js
```diff
@@ -970,6 +989,44 @@ class Searcher extends React.Component<SearcherDefaultProps, SearcherProps, Sear
...
   }
 
+  /**
+   * Add signal each time a filter is added/removed.
+   * When a filter is added, signal with weight 1 is created.
+   * When a filter is removed, signal with weight 0 is created.
+   */
+  addFacetFilterSignal(facetFilter: FacetFilter, weight: number) {
+    const queryDocuments = this.state.response ? this.state.response.documents : null;
+    const querySignal = queryDocuments ? queryDocuments[0].signal : null;
```
luigivaldez: Just double checking... is it ever possible to not have any documents in the results when we get here? I don't think so, right? Any facet filters you can apply will always lead to having at least one document in the results I think.
In that case, we should be fine here.

src/components/Searcher.js
```diff
@@ -970,6 +989,44 @@ class Searcher extends React.Component<SearcherDefaultProps, SearcherProps, Sear
...
+    signal.query = querySignal.query;
+    signal.queryTimestamp = querySignal.queryTimestamp;
+    signal.relevancyModelName = querySignal.relevancyModelName;
+    signal.relevancyModelNames = querySignal.relevancyModelNames;
+    signal.relevancyModelVersion = querySignal.relevancyModelVersion;
+    signal.signalTimestamp = Date.now();
+    signal.ttl = false;
+
+    const doc = new SearchDocument(new Map(), signal);
+    new Signals(this.props.baseUri).addSignal(doc, 'facet', weight);
```
luigivaldez: We should maybe refactor the `addSignal` method so it takes a signal instead of a document; it seems silly to have to create a document that the method doesn't really care about.
If we do this, we should make sure it is documented as a breaking change, but I think that's fine considering that the React 16 change is in here too... (I doubt anybody is calling this from their application code but better safe than sorry...)

luigivaldez: Actually, if we're going to make a breaking change, we should probably make the method static and pass the baseUri to it as a parameterl so you don't have to do the `new Signals()` part... just call `Signals.addSignal(baseUri, signal, type, weight)`.
How does this sound?

src/components/Searcher.js
shah-tejas: Yes, this is actually dependent on the signal of type click created. If that doesn't work for some reason, this would not too!

src/components/Searcher.js
shah-tejas: Sounds good. We could in fact create a new function that accepts the Signal object directly. This would ensure the original `addSignal()` still works. For signals of types `like` and `click', the expects to send a `SearchDocument` from which the `Signal` object would be extracted. I think we should still support the existing `addSignal()` which is used for these signal types, and maybe any other in the future as well?

src/components/Searcher.js
luigivaldez: If the user has somehow disabled metadata, than that's fine... we won't have signals for clicks and we won't have them for other things like facet filters.
I was just making sure that we always have at least one document in the search results...

src/components/Searcher.js
luigivaldez: I just took a quick peek in the Attivio GitHub and there are actually a couple of cases where people outside of SUIT are calling this method... so let's leave it intact for now...

And just add a second method... maybe called `addRawSignal` that takes just the signal object and no type or weight. The places where you're currently creating an empty SearchDocument in order to add a signal should call this new method instead...

PR #150, opened 2019-08-02
bug/PLAT-44229 : Ensure pointers in GeoMaps point to the location correctly

shah-tejas:
[PLAT-44229](https://jira.attivio.com/browse/PLAT-44229) : Incorrect location pointer in Map facet in searchui

GeoMaps in SearchUI points to the location [0,0] even though the pointer has the correct co-ordinates for locations. This results in the map plotting only 1 point at [0,0] which is incorrect. This bug was introduced in 5.6.1 as `SearchFacetBucket.value` returned from the backend API has the co-ordinates in string format `"<longitude>,<latitude>"`. However, UI expects value to be an object in the form of `{longitude: <longitude_value>, latitude: <latitude_value>}`.

Thus, the fix for this issue is to check if `SearchFacetBucket.value` is a string or an object, before passing it to the Mapbox component.

### DIFF
--- a/src/components/MapFacetContents.js
+++ b/src/components/MapFacetContents.js
@@ -218,11 +218,11 @@ class MapFacetContents extends React.Component<MapFacetContentsDefaultProps, Map
         // Keep track of the boundaries of the coordinates
         return (
           <Marker
-            coordinates={[longitude || 0, latitude || 0]}
+            coordinates={[longitude, latitude]}
             onClick={() => {
               this.props.addFacetFilter(bucket);
             }}
-            key={`${longitude || 0},${latitude || 0}`}
+            key={`${longitude},${latitude}`}
             style={{ cursor: 'pointer' }}
           >
             <Glyphicon glyph="map-marker" style={{ fontSize: '18px', color: '#2a689c' }} />
--- a/src/components/MapFacetContents.js
+++ b/src/components/MapFacetContents.js
@@ -200,15 +200,20 @@ class MapFacetContents extends React.Component<MapFacetContentsDefaultProps, Map
 
       const points = this.props.buckets.map((bucket) => {
         const value = bucket.value; // JSON.parse(bucket.value);
-        let longitude = 0;
-        let latitude = 0;
+        let longitude = NaN;
+        let latitude = NaN;
         if (typeof value === 'string') {
           const valueArr = value.split(',');
-          longitude = valueArr[0];
-          latitude = valueArr[1];
+          if (valueArr.length === 2) {
+            longitude = Number.parseFloat(valueArr[0]);
+            latitude = Number.parseFloat(valueArr[1]);
+          }
         } else {
           longitude = value.longitude;
           latitude = value.latitude;
+        }
+        if (isNaN(longitude) || isNaN(latitude)) {
+          return null;
         }
         // Keep track of the boundaries of the coordinates
         return (
--- a/src/components/MapFacetContents.js
+++ b/src/components/MapFacetContents.js
@@ -199,7 +199,7 @@ class MapFacetContents extends React.Component<MapFacetContentsDefaultProps, Map
       });
 
       const points = this.props.buckets.map((bucket) => {
-        const value = bucket.value; // JSON.parse(bucket.value);
+        const value = bucket.value;
         let longitude = NaN;
         let latitude = NaN;
         if (typeof value === 'string') {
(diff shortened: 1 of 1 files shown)


src/components/MapFacetContents.js
```diff
@@ -200,14 +200,24 @@ class MapFacetContents extends React.Component<MapFacetContentsDefaultProps, Map
 
       const points = this.props.buckets.map((bucket) => {
         const value = bucket.value; // JSON.parse(bucket.value);
+        let longitude = 0;
```
luigivaldez: If the values are not set in the bucket, then I don't know if it makes sense to default to (0,0).... I'd prefer to not show the point at all and maybe do a console.warn() to say that something wasn't right... (For geographic facets, the values should always be there, except in cases like the one that caused this bug where the data structure returned by the back end changed...

src/components/MapFacetContents.js
```diff
@@ -200,14 +200,24 @@ class MapFacetContents extends React.Component<MapFacetContentsDefaultProps, Map
...
+        let latitude = 0;
+        if (typeof value === 'string') {
+          const valueArr = value.split(',');
+          longitude = valueArr[0];
```
luigivaldez: These will the the values as strings, not numbers, won't they? You should be able to call Number.parseFloat() to convert them... since the Marker component expects the coordinates property to be of type `Array<number>`.

luigivaldez: Also, you should make sure that there are 2 items in the array you get from `split()`... if there's some random string without a comma, you'll get an error trying to access `valueArr[1]`.

src/components/MapFacetContents.js
```diff
@@ -200,14 +200,24 @@ class MapFacetContents extends React.Component<MapFacetContentsDefaultProps, Map
...
+          latitude = valueArr[1];
+        } else {
+          longitude = value.longitude;
+          latitude = value.latitude;
+        }
         // Keep track of the boundaries of the coordinates
         return (
           <Marker
-            coordinates={[value.longitude || 0, value.latitude || 0]}
+            coordinates={[longitude || 0, latitude || 0]}
```
luigivaldez: If you're already validating that the latitude and longitude are valid, you shouldn't need the `|| 0` here...

src/components/MapFacetContents.js
shah-tejas: Makes sense. But wouldn't it be incorrect to dismiss a location if it is [0,0]? Even that is a valid location co-ordinate actually!
Should we maybe default the co-ordinates to `NaN` and dismiss them if they are still `NaN` after assignment from value?

luigivaldez: If it is really (0, 0) we should definitely show it...
If the value for the bucket was "foo" I don't think we should show it...
So, yeah, the NaN thing would work. And Number.parseFloat() will return NaN if the string isn't a number, so that's good.

I'm not sure what React will do if the `points` array contains some null values. If that works and React just ignores them, like if you put `{null}` somewhere in your JSX, then that's fine. Otherwise, you might need to use `forEach()` instead of `map()` and only push the <Marker> components onto the `points` array, skipping the ones that are invalid.. (In actuality, I would expect that if one is invalid, they'll all be invalid, but it still can't hurt..)

shah-tejas: `points` is directly injected as child of `<Map>` and react seems to ignore `null` values in `points`. So, the `NaN` logic should work out :+1:

src/components/MapFacetContents.js
```diff
@@ -200,14 +200,29 @@ class MapFacetContents extends React.Component<MapFacetContentsDefaultProps, Map
...
+          const valueArr = value.split(',');
+          if (valueArr.length === 2) {
+            longitude = Number.parseFloat(valueArr[0]);
+            latitude = Number.parseFloat(valueArr[1]);
+          }
+        } else {
+          longitude = value.longitude;
+          latitude = value.latitude;
+        }
+        if (isNaN(longitude) || isNaN(latitude)) {
```
luigivaldez: You should call `Number.isNaN()` etc. because ESLint will complain about using the global function `isNaN()`.

src/components/MapFacetContents.js
```diff
@@ -200,14 +200,29 @@ class MapFacetContents extends React.Component<MapFacetContentsDefaultProps, Map
 
       const points = this.props.buckets.map((bucket) => {
         const value = bucket.value; // JSON.parse(bucket.value);
```
luigivaldez: We should probably delete this old comment as long as we're changing the file...

PR #152, opened 2019-08-07
bug/PLAT-40591 : Do not render facet filters that are already applied to search

shah-tejas:
[PLAT-40591](https://jira.attivio.com/browse/PLAT-40591) : When a facet is applied, that facet should not be rendered

While applying filters in searchui, following changes in the behavior of Facets are applied:
- A facet now allows to add multiple filters based on buckets of same Facet.
- If a particular facet bucket is applied as a filter, then that bucket won't be shown in the Facet (except Facts of type 'tagcloud' and 'timeseries').
- If a facet has buckets where all the buckets are used as filters, the facet won't be displayed.

### DIFF
--- a/src/components/FacetResults.js
+++ b/src/components/FacetResults.js
@@ -141,15 +141,10 @@ export default class FacetResults extends React.Component<FacetResultsDefaultPro
    * Filter Buckets in Facet that are already applied to the search filters.
    * Buckets of type tagcloud and timeseries do not need to be filtered.
    */
-  filterFacetBuckets(facet: SearchFacet, type: string): SearchFacet {
+  filterFacetBuckets = (facet: SearchFacet, type: string, facetFiltersMap: Map<string, FacetFilter>): SearchFacet => {
     if (type === 'tagcloud' || type === 'timeseries') {
       return facet;
     }
-    const facetFilters = this.context.searcher.state.facetFilters;
-    const facetFiltersMap: Map<string, FacetFilter> = new Map();
-    facetFilters.forEach((facetFilter: FacetFilter) => {
-      facetFiltersMap.set(facetFilter.filter, facetFilter);
-    });
     const filteredBuckets = [];
     facet.buckets.forEach((bucket: SearchFacetBucket) => {
       if (!facetFiltersMap.get(bucket.filter)) {
@@ -167,12 +162,17 @@ export default class FacetResults extends React.Component<FacetResultsDefaultPro
       facets.forEach((facet: SearchFacet) => {
         facetsMap.set(facet.name, facet);
       });
+      const facetFilters = this.context.searcher.state.facetFilters;
+      const facetFiltersMap: Map<string, FacetFilter> = new Map();
+      facetFilters.forEach((facetFilter: FacetFilter) => {
+        facetFiltersMap.set(facetFilter.filter, facetFilter);
+      });
       const results = [];
       this.props.orderHint.forEach((facetName) => {
         const facet = facetsMap.get(facetName);
         if (facet && this.shouldShow(facet)) {
           const type = this.getFacetDisplayType(facet.field);
-          const filteredFacet = this.filterFacetBuckets(facet, type);
+          const filteredFacet = this.filterFacetBuckets(facet, type, facetFiltersMap);
           results.push(<Facet
             facet={filteredFacet}
             type={type}
@@ -187,7 +187,7 @@ export default class FacetResults extends React.Component<FacetResultsDefaultPro
         if (!this.props.orderHint.includes(facet.name)) {
           if (this.shouldShow(facet)) {
             const type = this.getFacetDisplayType(facet.field);
-            const filteredFacet = this.filterFacetBuckets(facet, type);
+            const filteredFacet = this.filterFacetBuckets(facet, type, facetFiltersMap);
             results.push(<Facet
               facet={filteredFacet}
               type={type}
--- a/src/components/FacetResults.js
+++ b/src/components/FacetResults.js
@@ -4,6 +4,8 @@ import PropTypes from 'prop-types';
 
 import Facet from './Facet';
 import SearchFacet from '../api/SearchFacet';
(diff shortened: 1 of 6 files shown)


src/components/Searcher.js
```diff
@@ -930,22 +930,22 @@ class Searcher extends React.Component<SearcherDefaultProps, SearcherProps, Sear
...
-   * new one.
+   * same facet name already exists, it will add the new filter
+   * instead of replacing.
    */
   addFacetFilter(facetName: string, bucketLabel: string, filter: string) {
+    const updatedFacetFilters = this.state.facetFilters ? this.state.facetFilters : [];
+    const filterAlreadyExisting = updatedFacetFilters.some((facetFilter: FacetFilter) => {
+      return facetFilter.filter === filter;
+    });
+    if (filterAlreadyExisting) {
```
luigivaldez: Is this just to handle the case for the tag cloud where the user can potentially click the same value over and over? If so, maybe add a comment there...

Although, I think it would be better to make sure that the values in the tag cloud for already applied filters shouldn't be clickable. You'd need to do this in the `TagCloudFacetContents` component, I think... (I think Dorit mentions something like this in the ticket.)

src/components/FacetResults.js
```diff
@@ -135,6 +137,28 @@ export default class FacetResults extends React.Component<FacetResultsDefaultPro
...
+  /**
+   * Filter Buckets in Facet that are already applied to the search filters.
+   * Buckets of type tagcloud and timeseries do not need to be filtered.
+   */
+  filterFacetBuckets(facet: SearchFacet, type: string): SearchFacet {
+    if (type === 'tagcloud' || type === 'timeseries') {
+      return facet;
+    }
+    const facetFilters = this.context.searcher.state.facetFilters;
+    const facetFiltersMap: Map<string, FacetFilter> = new Map();
```
luigivaldez: It's probably not _super_ expensive to build this map for every facet we look at, but it might make sense to build it outside of here (e.g. in the `renderFacets()` method) and pass it in as a parameter to `filterFacetBuckets()`.

src/components/Searcher.js
shah-tejas: Do you mean that tag cloud buckets that are actually applied as filters should render as `<span>` instead of `<a>`?
If so, logic for `<TagCloud>` would probably need to be changed. As it does not currently have a mechanism to make certain buckets non-clickable.

luigivaldez: Yes... that's what I mean. They would still show up with the correct size and all, but they:
1. shouldn't be clickable (if changing the <a> to <span> is the best way to do that, that's great)
2. shouldn't do the scaling thing that Nikhil added (where when you hover over them they grow)

It may be that the change for #1 will make #2 happen automatically, depending on how the CSS is set up.

The <TagCloud> component already has a mechanism for making the whole thing not clickable (the `noLink` parameter). You'll need to add a flag like that (I suppose using the same name would make things simpler) to the `TagCloudValue` class (the default value of which should be false). Then you'll also need to modify the TagCloudFacetContents to find out which ones, if any, are already applied.

Does that makes sense?

shah-tejas: Yeah makes sense. I had the same approach in my mind! You described it far better :)

So, I have added a `noLink` property in the `TagCloudValue` class which will be set from `TagCloudFacetContents` if the search filter is applied for that tag cloud. In `<TagCloud>`, the tag is rendered as a `<span>` if `noLink` is set for that tag.

However, regarding scaling of the tag clouds, it is controlled in the CSS and not in complete control by suit. The scaling CSS property is added on the CSS class `attivio-cloud` which is the class for the parent `<ul>` for all child `<li>` tag values.
So, a solution could be to add a new class `attivio-cloud-noLink` added to `<li>` tags which have `noLink = true` which will prevent the scaling. This would ensure existing behavior stays intact as well...

luigivaldez: That sounds like a the way to go. You'll be able to prevent both the scaling and the color change that happen now when hovering, while still preserving the font size, etc., so they still display at the right weight according to the count for that bucket.

src/components/TagCloud.js
```diff
@@ -7,10 +7,13 @@ export class TagCloudValue {
   label: string;
   /** The value for this item. */
   value: number;
+  /** Whether the tag should be clickable */
+  noLink: boolean;
 
-  constructor(label: string, value: number) {
+  constructor(label: string, value: number, noLink: boolean) {
```
luigivaldez: I would add an explicit default value of `false` to the noLink parameter of the constructor. This will help people know that they don't have to pass this (in case they're not doing so now) and it'll make it explicit that not doing to will create a clickable tag item.

src/components/TagCloud.js
```diff
@@ -134,7 +137,8 @@ export default class TagCloud extends React.Component<TagCloudDefaultProps, TagC
       };
       const className = TagCloud.getClassNameForLevel(size);
       return (
-        this.props.noLink ? (<li key={tcv.label}>
+        (this.props.noLink || tcv.noLink) ?
+        (<li key={tcv.label} className="attivio-cloud-noLink">
```
luigivaldez: Please reformat this so the opening parenthesis wrapping the JSX isn't on the same line as the JSX (e.g. the opening `<li>` tag). The updated ESLint will complain about this and it'll be easier to fix it now... The same is true for the "else" part. (You'll be able to put the condition on the same line as the `return` statement, to make it more compact.... and the extra set of parentheses that are around the whole return value for someone reason. :-) )
```
return (this.props.noLink || tcv.noLink) ? (
  <li key={tcv.label} className="attivio-cloud-noLink">
    <span className={className}>
      {tcv.label}
    </span>
  </li>
) : (
  <li key={tcv.label}>
    <a className={className} onClick={callback} role="button" tabIndex={0}>
      {tcv.label}
    </a>
  </li>
);
```

shah-tejas:
Created PR in searchui for styling of non-clickable items - https://github.com/attivio/searchui/pull/88

PR #155, opened 2019-08-13
Task/plat 44218 : Add signals to promotions

shah-tejas:
[PLAT-44218](https://jira.attivio.com/browse/PLAT-44218) : Add signals to promotions

This PR is to add signal when a promotion (link or image) is clicked. Additionally, for other types of signals being added, `signal.ttl` is removed from the request to ensure it is defaulted to `true` from backend.
Also, the creation of signal for facet filter is refactored to ensure `-add` or `-remove` is suffixed to `signal.docId` when a facet is added or removed.

### DIFF
--- a/src/components/PlacementResult.js
+++ b/src/components/PlacementResult.js
@@ -46,7 +46,7 @@ export default class PlacementResult extends React.Component<PlacementResultDefa
     if (!addPromotionSignal) {
       return;
     }
-    const signalDocId = imageUrl ? `${imageUrl}-${linkUrl}` : `${linkText}-${linkUrl}`;
+    const signalDocId = imageUrl ? `${imageUrl}-${linkUrl || ''}` : `${linkText || ''}-${linkUrl || ''}`;
     const signalDocOrdinal = imageUrl ? 2 : 1;
     addPromotionSignal(signalDocId, signalDocOrdinal);
   }
--- a/src/components/PlacementResult.js
+++ b/src/components/PlacementResult.js
@@ -1,5 +1,6 @@
 // @flow
 import React from 'react';
+import PropTypes from 'prop-types';
 
 import Card from './Card';
 
@@ -33,6 +34,22 @@ export default class PlacementResult extends React.Component<PlacementResultDefa
   }
 
   static displayName = 'PlacementResult';
+
+  static contextTypes = {
+    searcher: PropTypes.any,
+  };
+
+  onPromotionClick() {
+    const { linkUrl, linkText, imageUrl } = this.props;
+    const searcher = this.context.searcher;
+    const addPromotionSignal = searcher ? searcher.addPromotionSignal : null;
+    if (!addPromotionSignal) {
+      return;
+    }
+    const signalDocId = imageUrl ? `${imageUrl}-${linkUrl}` : `${linkText}-${linkUrl}`;
+    const signalDocOrdinal = imageUrl ? 2 : 1;
+    addPromotionSignal(signalDocId, signalDocOrdinal);
+  }
 
   renderMarkupIframe() {
     const getBlobURL = (code) => {
@@ -83,7 +100,7 @@ export default class PlacementResult extends React.Component<PlacementResultDefa
     return (
       <Card style={{ marginBottom: '10px' }}>
         {this.props.linkUrl ? (
-          <a href={this.props.linkUrl} >
+          <a href={this.props.linkUrl} onClick={() => this.onPromotionClick()} >
             {contents}
           </a>
         ) : contents}
--- a/src/components/PlacementResult.js
+++ b/src/components/PlacementResult.js
@@ -39,16 +39,19 @@ export default class PlacementResult extends React.Component<PlacementResultDefa
     searcher: PropTypes.any,
   };
 
-  onPromotionClick() {
+  onPromotionClick = () => {
     const { linkUrl, linkText, imageUrl } = this.props;
     const searcher = this.context.searcher;
-    const addPromotionSignal = searcher ? searcher.addPromotionSignal : null;
-    if (!addPromotionSignal) {
+    if (!searcher) {
       return;
     }
-    const signalDocId = imageUrl ? `${imageUrl}-${linkUrl || ''}` : `${linkText || ''}-${linkUrl || ''}`;
+    // The docId for the signal is of the format:
+    // '{Display Text} - {Destination Link}' -- if the promotion is Link
+    // '{Image URL} - {Destination Link}' -- if the promotion is Image
+    // The docOrdinal for link promotion is 1 and image promotion is 2
+    const signalDocId = imageUrl ? `${imageUrl} - ${linkUrl || ''}` : `${linkText || ''} - ${linkUrl || ''}`;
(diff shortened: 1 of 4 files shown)


src/components/PlacementResult.js
```diff
@@ -34,6 +35,22 @@ export default class PlacementResult extends React.Component<PlacementResultDefa
...
+  };
+
+  onPromotionClick() {
+    const { linkUrl, linkText, imageUrl } = this.props;
+    const searcher = this.context.searcher;
+    const addPromotionSignal = searcher ? searcher.addPromotionSignal : null;
+    if (!addPromotionSignal) {
+      return;
+    }
+    const signalDocId = imageUrl ? `${imageUrl}-${linkUrl || ''}` : `${linkText || ''}-${linkUrl || ''}`;
```
luigivaldez: I'd add a comment here saying explicitly how you're building the "doc ID" from the promotion's properties. Anyone wanting to make use of these signals will need to know how to parse it to get those pieces back...

src/components/Searcher.js
```diff
@@ -994,8 +1031,9 @@ class Searcher extends React.Component<SearcherDefaultProps, SearcherProps, Sear
     if (!facet) {
       return;
     }
+    const docIDSuffix = (weight === 1) ? '-add' : '-remove';
```
luigivaldez: I would change the delimiter from "-" to something that's unlikely to appear in a facet filter like a pipe "|" character... This will make it easier to split later to get the filter string back...

src/components/Searcher.js
```diff
@@ -982,8 +1020,7 @@ class Searcher extends React.Component<SearcherDefaultProps, SearcherProps, Sear
    * When a filter is removed, signal with weight 0 is created.
    */
   addFacetFilterSignal(facetFilter: FacetFilter, weight: number) {
-    const queryDocuments = this.state.response ? this.state.response.documents : null;
-    const querySignal = queryDocuments && queryDocuments.length >= 1 ? queryDocuments[0].signal : null;
+    const querySignal = this.getDefaultQuerySignal();
```
luigivaldez: Again here I'd add a comment saying what you're going to be putting in the docId field of the signal.

We will want to add this to the documentation on Answers, too... if you can create a ticket in Jira to do that that says something like "Document the values in Signals for Facet Filters, Promotions, and AutoComplete" that'd be great — it really belongs in the signals documentation more than the Search UI docs....

src/components/Searcher.js
```diff
@@ -982,8 +1020,7 @@ class Searcher extends React.Component<SearcherDefaultProps, SearcherProps, Sear
    * When a filter is removed, signal with weight 0 is created.
    */
   addFacetFilterSignal(facetFilter: FacetFilter, weight: number) {
```
luigivaldez: I would change the name of the parameter from `weight` to `added` or `removed` and make it be Boolean. Then you can set the weight on the signal object accordingly and it'll be semantically clearer.

src/components/Searcher.js
```diff
@@ -1008,13 +1046,38 @@ class Searcher extends React.Component<SearcherDefaultProps, SearcherProps, Sear
...
 
+  /**
+   * Add signal each time a promotion is clicked.
+   */
+  addPromotionSignal = (signalDocID: string, signalDocOrdinal: number) => {
+    const querySignal = this.getDefaultQuerySignal(true);
+    if (!querySignal) {
+      return;
+    }
+    const signal = new SignalData();
```
luigivaldez: This seems like it would be error prone. If the SignalData object is ever changed to add additional fields, they won't be copied here. I would make a `clone()` method in the SignalData class and call it instead, and then just set the values you need to override. (Anyone editing the SignalData class will still need to update the `clone()` method but that will be expected, without having to search for all the uses of SignalData in the code.)

src/components/Searcher.js
```diff
@@ -495,6 +498,37 @@ class Searcher extends React.Component<SearcherDefaultProps, SearcherProps, Sear
...
+    }
+    if (createNewIfNotExisting) {
+      const savedUser = AuthUtils.getSavedUser();
+      const defaultSignalData = new SignalData();
+      if (savedUser) {
+        defaultSignalData.locale = 'en';
+        defaultSignalData.principal = `${AuthUtils.config.ALL.defaultRealm}:${savedUser.fullName}:${savedUser.userId}`;
+        defaultSignalData.query = this.state.query;
+        defaultSignalData.queryTimestamp = this.state.queryTimestamp;
+        defaultSignalData.relevancyModelName = 'default';
```
luigivaldez: The relevancy model used for the query is also in the searcher's state... we should try to get it from there instead of using "default" since customers will expect this to be correct.

src/components/Searcher.js
shah-tejas: Created https://jira.attivio.com/browse/PLAT-44324 for the documentation.

src/components/Searcher.js
shah-tejas: Sounds good. However, when there are no search results returned, the `relevancyModels` in the state is not populated. So, I put a check to see if the value is present in the state. If not, I have defaulted it to "default".

luigivaldez: The property `relevancymodelnames` in the state should generally be the currently selected relevancy model(s) (well, usually just one) that are added to the query request when searching. If you look at the code in the `Searcher`'s `getQueryRequest()` method, you can refactor the bit that decides what to put for the "relevancyModelNames" REST parameter into a separate method and call it from both `getQueryRequest()` and `getDefaultQuerySignal()`.

shah-tejas:
Fixed the typos... sometimes dark theme of VSCode is deceiving :)

src/components/Searcher.js
```diff
@@ -399,6 +399,15 @@ class Searcher extends React.Component<SearcherDefaultProps, SearcherProps, Sear
     };
   }
 
+  getRelevancyModels(): string[] {
```
luigivaldez: For the type "array of strings," we should be using the parameterized Flow type `Array<string>`...

src/components/Searcher.js
```diff
@@ -515,14 +521,15 @@ class Searcher extends React.Component<SearcherDefaultProps, SearcherProps, Sear
...
+      const { query, queryTimestamp } = this.state;
+      const relevancyModels = this.getRelevancyModels();
       if (savedUser) {
         defaultSignalData.locale = 'en';
         defaultSignalData.principal = `${AuthUtils.config.ALL.defaultRealm}:${savedUser.fullName}:${savedUser.userId}`;
         defaultSignalData.query = query;
         defaultSignalData.queryTimestamp = queryTimestamp;
         defaultSignalData.relevancyModelName = relevancyModels[0] || 'default';
-        defaultSignalData.relevancyModelNames = relevancyModels && relevancyModels.length > 0 ? relevancyModels : ['default'];
+        defaultSignalData.relevancyModelNames = relevancyModels.length > 0 ? relevancyModels : ['default'];
```
luigivaldez: Thinking about this, I'm not sure that we want to default it to a model called "default." I think we should leave it out altogether if we don't have anything specific to put here. (I.e., set `relevancyModelName` to `''` and `relevancyModelNames` to `[]`.)

PR #158, opened 2019-09-09
feat/PLAT-44338 : Allow the mast head to be hidden if the url contains /no-mast

shah-tejas:
[PLAT-44338](https://jira.attivio.com/browse/PLAT-44338): Clicking Search should render SearchUI

Admin 2.0 requires Search UI to be rendered without the Attivio logo, application name and the logged in user info in the mast, when `/no-mast` path is included in the url.

To hide the above elements from the attivio mast, we intercept the URL in `<Searcher>` and set the `hideMast` var to true if the URL contains `/no-mast`. In `<Masthead>`, if `context.searcher.state.hideMast` is set, we render a minimal view of the masthead that does not contain the attivio logo, application name and the logged in user info.

Additionally, when the logo was removed from the mast, the height of the mast was suppressed. Currently, the height of the mast was derived from the css class `attivio-globalmast-logo` which is part of the logo element. Thus, to have the same height of the mast in the minimal view, class `attivio-globalmast-logo` is added the `<div>` in the container.

### DIFF
--- a/src/components/Masthead.js
+++ b/src/components/Masthead.js
@@ -244,6 +244,18 @@ class Masthead extends React.Component<MastheadDefaultProps, MastheadProps, Mast
       logoutFunction = null;
     }
 
+    const hideMast = this.context.searcher.state.hideMast || false;
+    if (hideMast) {
+      return (
+        <header className="attivio-globalmast attivio-minwidth">
+          <div className="attivio-container">
+            {this.props.children}
+            <div className="attivio-globalmast-spacer attivio-globalmast-logo" />
+          </div>
+        </header>
+      );
+    }
+
     return (
       <header className="attivio-globalmast attivio-minwidth">
         <div className="attivio-container">
--- a/src/components/Searcher.js
+++ b/src/components/Searcher.js
@@ -208,6 +208,7 @@ type SearcherState = {
   resultsOffset: number;
   debug: boolean;
   queryTimestamp: number;
+  hideMast: boolean;
 };
 
 /**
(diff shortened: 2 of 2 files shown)


luigivaldez:
I'm confused about a couple of things? I thought the masthead was going to be hidden entirely when Search UI was in an iframe. Are you going to have two nested mastheads now, one with the logo and user name and one with just the search box? Shouldn't the search box, etc. be outside of the masthead in the new case?

Secondly, I don't think it makes sense for this to be part of the searcher's state... It's not really state, so I would think it should be a property passed in from the parent. That said, I'm unclear why the searcher needs to know about it at all. What else are you changing based on this hideMast condition? Is there a different PR that changes other components?

shah-tejas:
Well it was not so straightforward to simply hide the masthead for the SearchUI, as the search box and the Results/Insights toggle would be lost with that as well. For the initial run, the masthead will be shown with the search box and the Results/Insights toggle. Eventually, the search box would be out of the masthead.

Tracking whether to hide the mast was the easiest to do in the searcher's state as Masthead has access to it and can simply render a different view if this property was set. `location` is already passed as a prop to `<Searcher>` which is all needed to determine whether `hideMast` should be set.

Currently, `hideMast` is only to render a minimal view of the masthead with only the children passed. Other PRs for other apps would simply not render Masthead itself I think...

luigivaldez:
Couldn't you just have a property on the masthead component instead? The SearchUISearchPage, etc., could then either set it or not, since it's really the place that knows if it should be included.

shah-tejas:
Yes that could be done too...

shah-tejas:
Looks like `<Masthead>` itself has `location` passed on as props. That can be directly used without changing anything in the `<Searcher>` or even SearchUI.

luigivaldez:
It could be.. but I don't think the masthead should know anything about specific URIs... I think it makes more sense to just have the `<Masthead>` component take a flag as a property (maybe it should be called "noLogo" or "simple" since it's not hidden... in either case, it should be a boolean that you can leave out by default).

shah-tejas:
Created https://github.com/attivio/suit/pull/160 since this is already merged.

PR #160, opened 2019-09-13
feat/PLAT-44338 : Allow the mast head to be hidden if the url contains /no-mast

shah-tejas:
[PLAT-44338](https://jira.attivio.com/browse/PLAT-44338): Clicking Search should render SearchUI

This is a continuation https://github.com/attivio/suit/pull/158

The approach to store `hideMast` in Searcher is changed to pass in `simple` prop to `<Masthead>` if `/no-mast` is present in the URL. This is to separate the logic to read URLs from suit and should be rather passed as a prop from searchui. Thus in this case, `simple` prop needs to be passed to `<Masthead>` if a simplified version of the Masthead is to be rendered.

Additionally, `<SearchResultTags>` provides a way to navigate to the 360 Page if "Show 360° View" button is clicked. The URL for the 360 page is hardcoded here and should be changed to `/no-mast-doc360` if the simple masthead needs to be rendered. Thus, a new optional `url360Page` prop is passed to `<SearchResults>` which contains the URL for the 360 Page to be rendered.

Similarly, the page for Results is hardcoded and should be passed as a prop from searchui in `<Doc360Breadcrumbs>`.

### DIFF
--- a/src/components/DebugSearchResult.js
+++ b/src/components/DebugSearchResult.js
@@ -26,6 +26,12 @@ type DebugSearchResultProps = {
   showTags: boolean;
   /** Whether star ratings should be shown in the UI or not. Defaults to true. */
   showRatings: boolean;
+  /**
+   * Optional.
+   * The Doc 360 page URL that should be rendered when the user clicks
+   * on "Show 360° View" button. This would default to '/doc360' if
+   * nothing provided.
+   */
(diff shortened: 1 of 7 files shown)


klk1010:
Closing this, since we decided on an alternate approach for searchui.

PR #161, opened 2019-09-17
bug/PLAT-44229 : Fix to ensure center of map is correctly calculated

shah-tejas:
[PLAT-44229 ](https://jira.attivio.com/browse/PLAT-44229) : Incorrect location pointer in Map facet in searchui

During testing of PLAT-44229, it was discovered that the center of the Map is not calculated correctly (https://jira.attivio.com/browse/PLAT-44229?focusedCommentId=481115&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-481115)

This bug was introduced in 5.6.1 as SearchFacetBucket.value returned from the backend API has the co-ordinates in string format `"<longitude>,<latitude>"`. However, UI expects value to be an object in the form of `{longitude: <longitude_value>, latitude: <latitude_value>}`.

Thus added support to convert the string value into the required format to fix this issue.

### DIFF
--- a/src/components/MapFacetContents.js
+++ b/src/components/MapFacetContents.js
@@ -63,9 +63,25 @@ class MapFacetContents extends React.Component<MapFacetContentsDefaultProps, Map
 
   static calcState(buckets: Array<SearchFacetBucket>, zoom: number,
     geoFilters: Array<string>, updating: string): MapFacetContentsState {
-    const center = PositionUtils.calcCenter(buckets.map((bucket) => {
-      return bucket.value;
-    }));
+    const center = PositionUtils.calcCenter(buckets.reduce((reduceResult, bucket) => {
+      const value = bucket.value;
+      let longitude = NaN;
+      let latitude = NaN;
+      if (typeof value === 'string') {
+        const valueArr = value.split(',');
+        if (valueArr.length === 2) {
+          longitude = Number.parseFloat(valueArr[0]);
+          latitude = Number.parseFloat(valueArr[1]);
+        }
+      } else {
+        longitude = value.longitude;
+        latitude = value.latitude;
+      }
+      if (!Number.isNaN(longitude) && !Number.isNaN(latitude)) {
+        reduceResult.push({ longitude, latitude });
+      }
+      return reduceResult;
+    }, []));
     return {
       latitude: center.latitude,
       longitude: center.longitude,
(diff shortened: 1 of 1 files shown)


src/components/MapFacetContents.js
```diff
@@ -63,9 +63,25 @@ class MapFacetContents extends React.Component<MapFacetContentsDefaultProps, Map
 
   static calcState(buckets: Array<SearchFacetBucket>, zoom: number,
     geoFilters: Array<string>, updating: string): MapFacetContentsState {
-    const center = PositionUtils.calcCenter(buckets.map((bucket) => {
-      return bucket.value;
-    }));
+    const center = PositionUtils.calcCenter(buckets.reduce((reduceResult, bucket) => {
```
luigivaldez: Also, add a few comments explaining what's going on...

shah-tejas:
Addressed code review comments.

src/components/MapFacetContents.js
```diff
@@ -199,22 +229,12 @@ class MapFacetContents extends React.Component<MapFacetContentsDefaultProps, Map
...
-            latitude = Number.parseFloat(valueArr[1]);
-          }
-        } else {
-          longitude = value.longitude;
-          latitude = value.latitude;
-        }
-        if (Number.isNaN(longitude) || Number.isNaN(latitude)) {
+        // Return null if getCoordinatesFromBucket() returns null value
+        const coordinates = MapFacetContents.getCoordinatesFromBucket(bucket);
+        if (!coordinates) {
```
luigivaldez: Actually, this is essentially doing the same mapping and filtering as the boundary code, isn't it... maybe make the `getCoordinatesFromBucket()` function take the array off buckets and return an array of the non-null coordinates (probably renaming it to `getCoordinatesFromBuckets()`)... Then, here, you can just map that list to `<Maker>` components.

src/components/MapFacetContents.js
```diff
@@ -75,6 +78,33 @@ class MapFacetContents extends React.Component<MapFacetContentsDefaultProps, Map
     };
   }
 
+  /**
+   * Fetches the latitude and longitude from the SearchFacetBucket value.
```
luigivaldez: I would change "fetches" to "gets" or something since "fetch" in JavaScript implies a network call...

src/components/MapFacetContents.js
shah-tejas: This sounds like a good approach. Although, the `<Marker>` component also requires the corresponding `bucket` for the current coordinates.
In case the new function returns an array of coordinates, we will have to again get the `bucket` for the respective coordinate when we map that array.

luigivaldez: You're totally right. I wasn't thinking about the need for the bucket to apply the filter.

PR #162, opened 2019-09-18
bug/PLAT-40591

shah-tejas:
[PLAT-40591](https://jira.attivio.com/browse/PLAT-40591)

Handle the case ([described here](https://jira.attivio.com/browse/PLAT-40591?focusedCommentId=481143&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-481143)) where facets of same timeseries are applied to the filter even though they already exist.

Added the logic to check if facet is already added to the filter before adding a new facet filter.


shah-tejas:
Makes sense to replace the old timeseries filter with new filter as the new one can have different date range values. Thus, added a check in `addTimeSeriesFilter()` to remove existing timeseries filter before adding the new filter.
Also, added a check to ensure the same filter is not added back again.

PR #166, opened 2019-10-01
bug/PLAT-40591

shah-tejas:
[PLAT-40591](https://jira.attivio.com/browse/PLAT-40591)

This is a continuation of https://github.com/attivio/suit/pull/162. It fixes the issue reported [here](https://jira.attivio.com/browse/PLAT-40591?focusedCommentId=483200&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-483200), which is as follows:

On the Insights Page of SearchUI, it allows to add same facet bucket in the filter multiple times.

Additionally there is an issue with "Score" / "Classification Score" type facets where the application breaks when a score is clicked from the facet on either the Search Page or the Insights Page. This issue is present in SearchUI 1.0.7 as well but there the filter is not applied and the application does not break as well.
This issue is caused here:
https://github.com/attivio/suit/blob/8de1c97c8fa39be33d9021a413aac8aa14da63c6/src/components/NavbarFilter.js#L40
When original is of type number and not string in this case, the error `original.includes is not a function` is reported.
To ensure, original is a string, in SearchFacetBucket.displayLabel() the value of min is returned as string.


src/components/FacetInsights.js
```diff
@@ -124,14 +126,32 @@ export default class FacetInsights extends React.Component<FacetInsightsDefaultP
...
+    if (type === 'tagcloud' || type === 'timeseries') {
+      return facet;
+    }
+    const filteredBuckets = [];
+    facet.buckets.forEach((bucket: SearchFacetBucket) => {
+      if (!facetFiltersMap.get(bucket.filter)) {
+        filteredBuckets.push(bucket);
+      }
+    });
+    return Object.assign(facet, { buckets: filteredBuckets });
```
klk1010: Instead of `Object.assign`
```suggestion
    return { 
      ...facet, // spread operator
      buckets: filteredBuckets, // order matters - This will overwrite facet.buckets
    });
```

src/components/FacetInsights.js
```diff
@@ -124,14 +126,32 @@ export default class FacetInsights extends React.Component<FacetInsightsDefaultP
...
-  facetForField(fieldName: string, facetMap: Map<string, SearchFacet>) {
+  /**
+   * Filter Buckets in Facet that are already applied to the search filters.
+   * Buckets of type tagcloud and timeseries do not need to be filtered.
+   */
+  filterFacetBuckets = (facet: SearchFacet, type: string, facetFiltersMap: Map<string, FacetFilter>): SearchFacet => {
+    if (type === 'tagcloud' || type === 'timeseries') {
+      return facet;
+    }
+    const filteredBuckets = [];
```
klk1010: Instead of `forEach`, [reduce](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/reduce) works well for this.
```suggestion
    const filteredBuckets = facet.buckets.reduce((accumBuckets, currentBucket) => {
       return !facetFiltersMap.get(bucket.filter)
         ? [...accumBuckets, currentBucket]
         : accumBuckets; 
    }, []);
```

src/api/SearchFacetBucket.js
```diff
@@ -78,7 +78,7 @@ export default class SearchFacetBucket {
       if (this.max) {
         result = `${this.min} - ${this.max}`;
       }
-      result = this.min;
+      result = `${this.min}`;
```
luigivaldez: 👍 The return type in the signature of the function is "string" so this should probably already have been like this.

src/components/FacetInsights.js
shah-tejas: Not sure why, but flow seems to complain about this syntax.

```
return {
142: ...facet,
143: buckets: filteredBuckets,
144: };
^ object literal. This type is incompatible with the expected return type of
132: filterFacetBuckets = (facet: SearchFacet, type: string, facetFiltersMap: Map<string, FacetFilter>): SearchFacet => {
^^^^^^^^^^^ SearchFacet
```

klk1010: My example is returning an object: `{}`, while the flow declaration expects type `SearchFacet`.

klk1010: You'll need to either return a `SearchFacet` instead of an object or update the return type in the flow declaration.

shah-tejas: Addressed this by returning `new SearchFacet()`

src/components/FacetInsights.js
```diff
@@ -167,12 +185,18 @@ export default class FacetInsights extends React.Component<FacetInsightsDefaultP
     ObjectUtils.removeItem(facetOrder, 'keyphrases');
     ObjectUtils.removeItem(facetOrder, 'date');
 
+    const facetFilters = searcher.state.facetFilters;
```
klk1010: Prefer [destructuring](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment).
```suggestion
    const { facetFilters } = searcher.state;
```

PR #167, opened 2019-10-02
issue-113 Add share search

lipim12:
Picking up requested changes from https://github.com/attivio/suit/pull/114
Previously commits made by @nikhilj13


src/components/SearchBar.js
```diff
@@ -16,69 +16,87 @@ import Signals from '../api/Signals';
...
-  baseUri: string;
+  baseUri: string,
   /** The label to show on the search button. Defaults to "Go". */
-  buttonLabel: string;
+  buttonLabel: string,
   /** If set, this is the route to navigate to upon executing a search. By default, no navigation will occur when searching. */
-  route: string | null;
+  route: string | null,
+  /** Specifies if share search option should be displayed or not, false by default */
+  shareSearch: boolean,
```
klk1010: Indicate this is optional.
```suggestion
  shareSearch?: boolean,
```

src/components/SearchBar.js
```diff
@@ -16,69 +16,87 @@ import Signals from '../api/Signals';
...
   /**
    * If set, a new signal of this type would be added when an autocomplete item is selected.
    */
-  createAutoCompleteSignal?: boolean;
+  createAutoCompleteSignal?: boolean,
+  /**
+   * The message body for when a search is shared.
+   */
+  shareMessage: string,
```
klk1010: This needs to be optional for backwards compatibility.
```suggestion
  shareMessage?: string,
```

src/components/SearchBar.js
```diff
@@ -16,69 +16,87 @@ import Signals from '../api/Signals';
...
+  /**
+   * Subject of the email for when a search is shared.
+   */
+  subject: string,
```
klk1010: Same
```suggestion
  subject?: string,
```

src/components/SearchBar.js
```diff
@@ -16,69 +16,87 @@ import Signals from '../api/Signals';
...
+  /**
+   * Email address the search will be shared with when using shareSearch.
+   */
+  email: string,
```
klk1010: Same
```suggestion
  email?: string,
```

src/components/SearchBar.js
```diff
@@ -96,7 +114,12 @@ class SearchBar extends React.Component<SearchBarDefaultProps, SearchBarProps, S
     autoCompleteUri: null,
     route: null,
     baseUri: '',
+    shareSearch: false,
     createAutoCompleteSignal: false,
+    shareMessage: 'Hey,\n\nI think you would be interested in these search results that I found using the most'
```
klk1010: Prefer [template literals](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals). They're easier to read.
```suggestion
    shareMessage: `Hey, I think you would be interested in these search results that I found using Attivio, a leader in cognitive search and knowledge discovery. Here is the link:`,
```

package-lock.json
```diff
@@ -6751,12 +6751,20 @@
       }
     },
     "eslint-plugin-flowtype": {
-      "version": "2.35.1",
-      "resolved": "https://registry.npmjs.org/eslint-plugin-flowtype/-/eslint-plugin-flowtype-2.35.1.tgz",
-      "integrity": "sha512-YTCeVsMOi3ga8PJjdAV97FaTNXNknzrO+4ZDMHJN65i4uMjL5KgfgQZpyVsBirWOfgXMKRduxpsyM64K/0LbXw==",
+      "version": "4.3.0",
```
klk1010: I noticed you updated `<SearchBar />` to use the more up to date flow declarations to use commas instead of semicolons. Will all the existing files using the legacy semicolon declarations cause linting errors? PLAT-44570 updates all dependencies. If this causes errors when you run `npm run lint`, it might be better to wait and just use semicolons here for now.

src/components/SearchBar.js
```diff
@@ -16,69 +16,87 @@ import Signals from '../api/Signals';
 declare var webkitSpeechRecognition: any; // Prevent complaints about this not existing
 
 type SearchBarProps = {
-  history: PropTypes.object.isRequired;
+  history: PropTypes.object.isRequired,
   /** If set, this will be styled to live inside a Masthead component. */
-  inMasthead: boolean;
+  inMasthead: boolean,
```
klk1010: For reference, these booleans have defaults set and should also indicate they are optional, so they aren't a good reference for best practices. This one should be:
`inMasthead?: boolean,`

src/components/SearchBar.js
```diff
@@ -367,56 +412,81 @@ class SearchBar extends React.Component<SearchBarDefaultProps, SearchBarProps, S
...
-        <input
-          type="search"
-          className={inputClass}
-          placeholder={placeholder}
-          onChange={this.queryChanged}
-          onKeyDown={this.doKeyPress}
-          value={query}
-        />
-      );
+    const inputComponent = this.props.autoCompleteUri ? (
```
klk1010: This isn't part of your changes, but `inputComponent` should be moved out of the render function into its own render function. Same with `shareSearch`. Instead of reference the const you would call the render function directly.

src/components/SearchBar.js
```diff
@@ -367,56 +412,81 @@ class SearchBar extends React.Component<SearchBarDefaultProps, SearchBarProps, S
...
+          border: 'none',
+          background: 'transparent',
+          cursor: 'pointer',
+        }}
+      >
+        <Glyphicon onClick={this.shareSearch} glyph="share" style={{ color: 'white', fontSize: '1.1em' }} />
+      </span>
+    ) : (
+      ''
+    );
```
klk1010: Move into its own function:
```
renderShareSearch() {
  const { shareSearch } = this.props;
  return shareSearch && (
          <span
        className="attivio-smalltoolbar-btn"
        title="Share this search via email"
        style={{
          position: 'relative',
          top: '2px',
          left: '-1px',
          color: '#fff',
          border: 'none',
          background: 'transparent',
          cursor: 'pointer',
        }}
      >
        <Glyphicon onClick={this.shareSearch} glyph="share" style={{ color: 'white', fontSize: '1.1em' }} />
      </span>
  );
}
```

src/components/SearchBar.js
```diff
@@ -367,56 +412,81 @@ class SearchBar extends React.Component<SearchBarDefaultProps, SearchBarProps, S
...
+                this.submitButton = c;
+              }}
             >
               {this.props.buttonLabel}
             </button>
           </div>
           {suggestionList}
         </div>
         {languageControl}
+        {shareSearch}
```
klk1010: Then this would be `renderShareSearch()`

src/components/SearchBar.js
```diff
@@ -278,14 +301,26 @@ class SearchBar extends React.Component<SearchBarDefaultProps, SearchBarProps, S
     }
   }
 
+  shareSearch() {
+    const username = AuthUtils.getUserName(AuthUtils.getSavedUser());
+    const signature = `\nfrom \n${username}`;
+    const message = this.props.shareMessage;
+    const searchLink = window.location.href;
```
klk1010: Can you get this data from [`withRouter`](https://reacttraining.com/react-router/web/api/history)?

src/components/SearchBar.js
```diff
@@ -278,14 +301,26 @@ class SearchBar extends React.Component<SearchBarDefaultProps, SearchBarProps, S
...
+    const emailBody = message + searchLink + signature;
+    const subject = this.props.subject;
+    const emailAddress = this.props.email;
+    document.location.href = `mailto:${emailAddress}?subject=${encodeURIComponent(subject)}&body=${encodeURIComponent(emailBody)}`;
```
klk1010: Could you just use `<a />`, with this string as the `href` instead of having an onClick function?

klk1010 (changes requested):
Address feedback

luigivaldez (commented):
Hmm... it feels like having various optional pieces be part of the search bar component itself is Turing it into a sort of "Franken-component." Not just the sharing functionality, but also things like saved searches and other things that people are and have been adding, or even maybe the language popup menu...

For example this new sharing functionality should really be, I think, a separate component that can be included inside the `<SearchBar>` as a child. This would eliminate the need to add all of these new properties that don't really have anything to do with the search bar _per se_ (like the e-mail message to send, for example).

I was going to suggest waiting for a future time to refactor things, but then backward compatibility becomes a pain, so maybe we should try and do this now.

It should be pretty straightforward to do... Just move the share search stuff into its own component and include it as a child component when the `<SearchBar>` is used. Then, in the search bar's render method, you can just render the search bar's children where you currently render `{shareSearch}`.

Does this make sense?

The result would be that adding arbitrary new functionality to the search bar would become simpler in general.

klk1010:
@luigivaldez makes a good point. `<SearchBar />` is becoming unwieldy. Rather than just move the areas I suggested into a separate functions, creating new components is better. Including them as children would definitely simplify the logic.

lipim12:
It does make sense to move the specifics of share search into a new component. I'll refactor and make a new commit. Thanks!

src/components/SearchBar.js
```diff
@@ -367,56 +429,45 @@ class SearchBar extends React.Component<SearchBarDefaultProps, SearchBarProps, S
...
+                this.submitButton = c;
+              }}
             >
               {this.props.buttonLabel}
             </button>
           </div>
           {suggestionList}
         </div>
         {languageControl}
+        {shareSearch}
```
klk1010: This can just be:
`{this.props.allowShareSearch && <ShareSearch shareMessage={this.props.shareMessage} subject={subject} email={emailAddress} />}`

styleguide.config.js
```diff
@@ -14,11 +14,7 @@ const ourWebpackConfig = {
     rules: [
       {
         test: /\.less$/,
-        use: [
-          'style-loader',
-          'css-loader',
-          'less-loader',
-        ],
+        use: ['style-loader', 'css-loader', 'less-loader'],
```
klk1010: Revert Prettier changes to this file.

src/components/SearchResults.js
```diff
@@ -18,7 +18,7 @@ import SearchDocument from '../api/SearchDocument';
  * rendered by this function. If no SearchResultRenderer functions handle the
  * rendering, then the document will be rendered by the 'list' renderer.
  */
-export type SearchResultRenderer = (doc: SearchDocument, position: number, baseUri: string, key: string) => any;
+export type SearchResultRenderer = (doc: SearchDocument, position: number, baseUri: string, key: string, hide360Link?: boolean) => any;
```
klk1010: 😅 Thanks for catching this! I missed it when I added the hide link option.

src/components/SearchBar.js
```diff
@@ -367,56 +429,45 @@ class SearchBar extends React.Component<SearchBarDefaultProps, SearchBarProps, S
...
+        <div className={subContainerClass} role="search">
           <div className="form-group">
-            {inputComponent}
+            {this.renderInputComponent(query, language)}
             {showMicrophone ? (
               <a onClick={this.startSpeechRecognition} role="button" tabIndex={0}>
                 <span className="attivio-globalmast-search-mic-icon attivio-icon-microphone" style={micStyle} />
               </a>
-            ) : ''}
+            ) : (
```
klk1010: Revert Prettier changes

src/components/SearchBar.js
```diff
@@ -353,8 +416,7 @@ class SearchBar extends React.Component<SearchBarDefaultProps, SearchBarProps, S
             background: 'transparent',
           }}
         >
-          <Glyphicon glyph="search" style={{ color: 'white' }} />
-          {' '}
+          <Glyphicon glyph="search" style={{ color: 'white' }} />{' '}
```
klk1010: Revert Prettier changes

src/components/SearchBar.js
```diff
@@ -278,15 +303,41 @@ class SearchBar extends React.Component<SearchBarDefaultProps, SearchBarProps, S
...
-    const showMicrophone = this.props.allowVoice && ('webkitSpeechRecognition' in window);
+    const showMicrophone = this.props.allowVoice && 'webkitSpeechRecognition' in window;
     const micStyle = {};
     if (this.state.recognizing) {
       micStyle.backgroundSize = '125%';
     }
 
     const containerClass = this.props.inMasthead ? 'attivio-globalmast-search-container' : '';
-    const inputClass = this.props.inMasthead ? 'form-control attivio-globalmast-search-input' : 'form-control';
+    const subContainerClass = this.props.allowShareSearch ? 'attivio-globalmast-search-share-search' : 'attivio-globalmast-search';
```
klk1010: What if `inMasthead` is specified?

src/components/ShareSearch.js
```diff
@@ -0,0 +1,83 @@
...
+  };
+
+  static contextTypes = {
+    searcher: PropTypes.any,
+  };
+
+  static displayName = 'ShareSearch';
+
+  render() {
+    const username = AuthUtils.getUserName(AuthUtils.getSavedUser());
```
klk1010: Prefer [destructuring](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment):
>**Default values**
>A variable can be assigned a default, in the case that the value unpacked from the object is undefined.
>**Assigning to new variables names and providing default values**
>A property can be both 1) unpacked from an object and assigned to a variable with a different name and 2) assigned a default value in case the unpacked value is undefined.
```
const {
  shareMessage: message,
  history: { location: { pathname: searchlink } }, // rename here if you like
  subject = '', // set a default value
  email: emailAddress = '', // rename and set a default value
} = this.props;
const username = AuthUtils.getUserName(AuthUtils.getSavedUser());
const signature = `\nfrom \n${username}`;
const emailBody = message + searchLink + signature;
```

src/components/SearchBar.js
```diff
@@ -278,15 +303,41 @@ class SearchBar extends React.Component<SearchBarDefaultProps, SearchBarProps, S
     }
   }
 
+  renderInputComponent(query, language) {
```
klk1010: Prefer [destructuring assignment](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment) of props.
```
const {
  placeholder: defaultPlaceholder = '',
  placeholderAdvanced = '',
  allowLanguageSelect,
  inMasthead,
  autoCompleteUri,
  baseUri,
} = this.props;

const placeholder = allowLanguageSelect && language === 'advanced'
  ? placeholderAdvanced
  : defaultPlaceholder;
```

klk1010 (changes requested):
Address feedback

PR #168, opened 2019-10-04
feat/PLAT-44338: Add support for show360 prop to SearchResults

klk1010:
Adds support for `show360` prop to `<SearchResults />`. The prop is optional. If omitted, the 360 view link is omitted.
`<SearchResults show360 ... />` passes it down to:
`<DebugSearchResult />` and `<ListSearchResult />` who each pass it down to: `<SearchResultTags />` This is the component that decides whether or not to render the 360 link.
There is an existing `view360Label` prop that when specified as `null` results in the 360 link to not be shown. However, reusing this prop for my use case would make it overly complex. I would need to move the default text to a central location and export it, which external libraries wouldn't have access to, or else add a special "useDefault" string, or something similar. It's less complex to create a new prop that doesn't relate to the label. This approach omits the label if it's `null` or if `show360` is false.


klk1010:
Good point about negative versus positive. I'll switch this to `no360Link` and default it to false instead so it follows the standard pattern for boolean props.

This allows for omission of the link for any reason. Are there any other elements you think we should omit? If it's just this one, then I don't think suit should care about why I'm omitting it, just support the option, but if we have a category of items that navigate, more than just this one, then that approach makes a lot of sense.

As far as passing it all the way down - it's definitely not the prettiest approach. I think that's something we should consider as part of the suit refactor.

PR #169, opened 2019-10-10
issue-108: Add saved search feature

lipim12:
Addressing requested changes on https://github.com/attivio/suit/pull/110

A few notes about refactoring with MetadataManager component:
- It is made for handling metadata management. You can think of a few examples where an API request is required but it is not relevant to search or searcher. You can add such functions in MetadataManager and use it from any other component by referencing it as `this.context.metadataManager`. 
- Originally this idea about refactoring with MetadataManager has been discussed before with Nikhil, Kristi, Luis, and Brandon.
- Workflow, Zone, Table for Saved Searches have been hardcoded as of now, with an assumption that `.metadata` zone would become OOTB in upcoming releases.


lipim12:
> @lipim12, to confirm, is this PR intended to replace #110 ? If so, I'll go ahead and close #110

Yes

PR #170, opened 2019-10-11
issue-108: Add saved search feature

lipim12:
Closing https://github.com/attivio/suit/pull/169 and making a new PR from a branch on suit (and not my forked repo)

**Addressing requested changes on #110**

_A few notes about refactoring with `MetadataManager` api class:_

- It is made for handling metadata management. You can think of a few examples where a component needs to make an API request but it is not an authentic search to be put in searcher or search. You can add such functions in` MetadataManager`.

- Originally this idea about refactoring with MetadataManager has been discussed before with Nikhil, Kristi, Luis, and Brandon.

- Workflow, Zone, Table for Saved Searches have been hardcoded as of now, with an assumption that `.metadata` zone would become OOTB in upcoming releases.

_How to use `MetadataManager`:_

- Create an instance of `MetadataManager` in the constructor of a component where you need to use its services. Initialize it with Searcher.Search.

```
example: MetadataManager;

constructor(props: ThisComponentProps) {
  super(props);
  this.example = new MetadataManager(this.context.searcher.search);
}

```
- To use a function of `MetadataManager`, call it using the created instance of `MetadataManager` and pass in required arguments.

```
sampleFunction = () => {
  ...
  ...
  this.example.someFunction(arg0, arg1);
  ...
}

```

PR #175, opened 2019-10-18
Updates and Improvement for Comments 

lipim12:
**Picking up https://github.com/attivio/suit/pull/147 previously made by @nikhilj13**

**Addresses @luigivaldez 's comments on https://github.com/attivio/suit/pull/138**

Refactored with `MetadataManager` class in `api` directory.

**A few notes about refactoring with `MetadataManager ` `api ` class:**

- It is made for handling metadata management. You can think of a few examples where a component needs to make a complex API request but it is not an authentic search to be put in searcher or search. You can add such functions in `MetadataManager`.

- Originally this idea about refactoring with `MetadataManager `has been discussed before with Nikhil, Kristi, Luis, and Brandon.

- Workflow, Zone, Table for comments have been hardcoded as of now, with an assumption that `.metadata` zone would become OOTB in upcoming releases.

**How to use `MetadataManager`:**

- Create an instance of `MetadataManager `in the constructor of a component where you need to use its services. Initialize it with Searcher.Search.

`example: MetadataManager;`

```
constructor(props: ThisComponentProps) {
  super(props);
  this.example = new MetadataManager(this.context.searcher.search);
}
```

- To use a function of `MetadataManager`, call it using the created instance of `MetadataManager `and pass in required arguments.
```
sampleFunction = () => {
  ...
  ...
  this.example.someFunction(arg0, arg1);
  ...
}
```


lipim12:
https://github.com/attivio/suit/pull/170 also adds/refactors code using `MetadataManager`

PR #176, opened 2019-10-18
Feature/issue 174

brandon-bogan:
Add ability to use tabs to filter results


src/components/FacetResults.js
```diff
@@ -169,22 +173,24 @@ export default class FacetResults extends React.Component<FacetResultsDefaultPro
...
-          const filteredFacet = this.filterFacetBuckets(facet, type, facetFiltersMap);
-          results.push(<Facet
-            facet={filteredFacet}
-            type={type}
-            key={facet.name}
-            maxBuckets={this.props.maxFacetBuckets}
-            collapse
-            entityColors={this.props.entityColors}
-          />);
+        if (this.props.hiddenFacets.indexOf(facetName) === -1) {
```
luigivaldez, line 164: You can say `hiddenFacets.includes(facetName)` instead of the indexOf/-1 thing, which is a little clearer and easier to read.

src/components/FacetResults.js
```diff
@@ -44,6 +44,8 @@ type FacetResultsProps = {
    * buckets. By default, facets with no buckets will be hidden.
    */
   showEmptyFacets: boolean;
+  /** Facets to not render */
+  hiddenFacets: Array<string>;
```
luigivaldez, line 48: Is this part of the Tabs change? Is it here so that the facet being used for the tabs doesn't need to be rendered as a separate facet?

src/components/FacetResults.js
```diff
@@ -169,22 +173,24 @@ export default class FacetResults extends React.Component<FacetResultsDefaultPro
...
+              maxBuckets={this.props.maxFacetBuckets}
+              collapse
+              entityColors={this.props.entityColors}
+            />);
+          }
         }
       });
       facets.forEach((facet: SearchFacet) => {
-        if (!this.props.orderHint.includes(facet.name)) {
+        if (!this.props.orderHint.includes(facet.name) && this.props.hiddenFacets.indexOf(facet.name) === -1) {
```
luigivaldez, line 181: Ditto here.

src/components/SearchResults.js
```diff
@@ -44,44 +48,46 @@ type SearchResultsProps = {
...
-  showTags?: boolean;
-  /** Whether star ratings should be shown in the UI or not. Defaults to true. */
-  showRatings?: boolean;
   /**
    * Whether or not to show 360 view link on search result. Defaults to false.
    */
   hide360Link?: boolean;
+  /** Whether or not to show tabs based on a facet */
+  showTabs: boolean;
+  /** Facet field to use for tabs */
```
luigivaldez, line 57: Mention in the comment that the default field is "table."
Alternatively, it might make sense to not have two things to set and just require the user to set the tabsField property to the field they want. If they don't set the field, then there are no tabs; this way the `showTabs` property isn't needed and it's a little simpler.
I might even change the property name from `tabsField` to `tabsFacetField`... Then the comment could say "The facet field to use for splitting the list of search results into tabs. Must be a facet that's present in the query request. If not set, all search results are displayed in a single list."

I might even mention that the tabs here will correspond to the buckets returned for the given facet, just to make it explicit.

You might want to add a note that the user might want to also add the facet field listed here to the `hiddenFacets` property of the FacetResults component so they aren't in 2 places...

src/components/SearchResults.js
```diff
@@ -44,44 +48,46 @@ type SearchResultsProps = {
...
+  tabsField: string;
+  /** List of explicit tabs to always show. If empty, tab values will be dynamic based on query response. */
+  tabList: Array<string>;
```
luigivaldez, line 60: Maybe change the property name to something that includes the idea of them being always there... like `fixedTabs`?
Also, I think it should be clear that 1) this only applies if tabs are being requested at all and 2) that if there are more facet buckets than the items in this list, they will be shown in addition to these.

src/components/SearchResults.js
```diff
@@ -91,6 +97,94 @@ export default class SearchResults extends React.Component<SearchResultsDefaultP
 
   static displayName = 'SearchResults';
 
+  constructor(props: SearchResultsProps) {
+    super(props);
+    this.state = {
+      activeTableTabKey: 'All',
+    };
+
+    (this: any).wrapResultsInTabs = this.wrapResultsInTabs.bind(this);
```
luigivaldez, line 96: This binding call isn't necessary here...

src/components/SearchResults.js
```diff
@@ -91,6 +97,94 @@ export default class SearchResults extends React.Component<SearchResultsDefaultP
...
+    (this: any).handleTabChange = this.handleTabChange.bind(this);
+  }
+
+  wrapResultsInTabs(results: Array<any>, response: QueryResponse) {
+    const facets = response.facets;
+    let tables = [];
+    let tableFacet;
```
luigivaldez, line 103: Ditto.

src/components/SearchResults.js
```diff
@@ -91,6 +97,94 @@ export default class SearchResults extends React.Component<SearchResultsDefaultP
...
+      activeTableTabKey: 'All',
+    };
+
+    (this: any).wrapResultsInTabs = this.wrapResultsInTabs.bind(this);
+    (this: any).handleTabChange = this.handleTabChange.bind(this);
+  }
+
+  wrapResultsInTabs(results: Array<any>, response: QueryResponse) {
+    const facets = response.facets;
+    let tables = [];
```
luigivaldez, line 102: This isn't just for tables, right? Maybe make the name more generic?

src/components/SearchResults.js
```diff
@@ -91,6 +97,94 @@ export default class SearchResults extends React.Component<SearchResultsDefaultP
...
+      tables = tableFacet.buckets.map((bucket: SearchFacetBucket) => {
+        const innerContents = activeTableTabKey === bucket.filter ? results : '';
+        return (
+          <Tab eventKey={bucket.filter} title={`${bucket.value} (${bucket.count})`}>
+            {innerContents}
+          </Tab>
+        );
+      });
+    } else if (this.props.tabList.length > 1) {
+      tables = this.props.tabList.map((tableName: string) => {
```
luigivaldez, line 130: If there are documents that aren't in one of the buckets mentioned in the tabList property, do you just ignore them?

src/components/Searcher.js
```diff
@@ -900,12 +917,19 @@ class Searcher extends React.Component<SearcherDefaultProps, SearcherProps, Sear
   }
 
   /**
-   * Add a query filter (in AQL) to the query request.
+   * Add a geospactial query filter (in AQL) to the query request.
```
luigivaldez, line 903: geospatial?

src/components/Searcher.js
```diff
@@ -951,7 +975,7 @@ class Searcher extends React.Component<SearcherDefaultProps, SearcherProps, Sear
 
 
   /**
-   * Add multiple query filters (in AQL) to the query request.
+   * Add multiple geospacial query filters (in AQL) to the query request.
```
luigivaldez, line 961: geospatial

src/components/SearchResults.js
```diff
@@ -91,6 +97,94 @@ export default class SearchResults extends React.Component<SearchResultsDefaultP
...
+        tableFacet = f;
+      }
+    });
+
+    let { activeTableTabKey } = this.state;
+    if (!this.context.searcher.state.filters || this.context.searcher.state.filters.length <= 0) {
+      activeTableTabKey = 'All';
+    } else {
+      this.context.searcher.state.filters.forEach((f: string) => {
+        if (f.indexOf('table') > -1) {
```
luigivaldez, line 115: Shouldn't this be the field name instead of "table"?

src/components/SearchResults.js
```diff
@@ -91,6 +97,94 @@ export default class SearchResults extends React.Component<SearchResultsDefaultP
...
+
+    (this: any).wrapResultsInTabs = this.wrapResultsInTabs.bind(this);
+    (this: any).handleTabChange = this.handleTabChange.bind(this);
+  }
+
+  wrapResultsInTabs(results: Array<any>, response: QueryResponse) {
+    const facets = response.facets;
+    let tables = [];
+    let tableFacet;
+    facets.forEach((f: SearchFacet) => {
```
luigivaldez, line 104: I would use the `find()` method on the array class in JavaScript. It'll make the semantics of what you're doing more obvious.

luigivaldez: Also, if the tabsField isn't one of the facets in the search result, should you maybe just return `results` as-is instead of putting them all under an "All" tab?

src/components/SearchResults.js
```diff
@@ -91,6 +97,94 @@ export default class SearchResults extends React.Component<SearchResultsDefaultP
...
+      if (f.name === this.props.tabsField) {
+        tableFacet = f;
+      }
+    });
+
+    let { activeTableTabKey } = this.state;
+    if (!this.context.searcher.state.filters || this.context.searcher.state.filters.length <= 0) {
```
luigivaldez, line 111: Here. too, I'd do the positive case first... and leave the "All" case to the `else` part of the `if` statement.

src/components/SearchResults.js
```diff
@@ -91,6 +97,94 @@ export default class SearchResults extends React.Component<SearchResultsDefaultP
...
+      activeTableTabKey = 'All';
+    } else {
+      this.context.searcher.state.filters.forEach((f: string) => {
```
luigivaldez, line 114: Here too you can use find instead of forEach

luigivaldez: What happens if there is one or more filters but there's nothing that matches your field name?

src/components/SearchResults.js
```diff
@@ -91,6 +97,94 @@ export default class SearchResults extends React.Component<SearchResultsDefaultP
...
+        if (f.indexOf('table') > -1) {
+          activeTableTabKey = f;
+        }
+      });
+    }
+    if (this.props.tabList.length < 1 && tableFacet && tableFacet.buckets) {
```
luigivaldez, line 120: So... the semantics are this?
If the caller specifies a list of tabs (in the `tabList` property, then those are used exactly as they are there, regardless of what's in the facet result. If there is no `tabList` specified, then the values from the facet results are used instead. If that's the case, I think it ought maybe to be pointed out somewhere more explicitly.

src/components/SearchResults.js
```diff
@@ -91,6 +97,94 @@ export default class SearchResults extends React.Component<SearchResultsDefaultP
...
+      tables = tableFacet.buckets.map((bucket: SearchFacetBucket) => {
```
luigivaldez, line 121: If the `tabList` property is not set and you're adding the tabs baed on the values in the facet response, it will be in order of most to least populated as it is now, I think, right? I think it would be weird for users to have tabs appear in a different order depending on what they search for... I might vote for always alphabetizing the tabs in this case. (With "All" always as the first one.)

src/components/SearchResults.js
```diff
@@ -91,6 +97,94 @@ export default class SearchResults extends React.Component<SearchResultsDefaultP
...
+        const innerContents = activeTableTabKey === bucket.filter ? results : '';
+        return (
+          <Tab eventKey={bucket.filter} title={`${bucket.value} (${bucket.count})`}>
```
luigivaldez, line 124: Instead of bucket.value, you might want to use the SearchFacetBucket class' `displayLabel()` method which will (theoretically) do a better job at figuring out what the user wants to see...

src/components/SearchResults.js
```diff
@@ -91,6 +97,94 @@ export default class SearchResults extends React.Component<SearchResultsDefaultP
...
+        );
+      });
+    } else if (this.props.tabList.length > 1) {
+      tables = this.props.tabList.map((tableName: string) => {
+        let isDisabled = true;
+        let count = 0;
+        let innerContents;
+        let eventKey = tableName;
+        if (tableFacet && tableFacet.buckets) {
+          tableFacet.buckets.forEach((facet: SearchFacetBucket) => {
```
luigivaldez, line 136: Again, I'd use `find()` here.

src/components/SearchResults.js
```diff
@@ -1,13 +1,17 @@
 // @flow
 import React from 'react';
 import PropTypes from 'prop-types';
+import { Tabs, Tab } from 'react-bootstrap';
```
luigivaldez, line 4: Is there a reason to use the bootstrap <Tabs> and <Tab> directly instead of using a <TabPanel> component (the same thing that's used when there are child documents from a join...)?

src/components/SearchResults.js
```diff
@@ -91,6 +97,94 @@ export default class SearchResults extends React.Component<SearchResultsDefaultP
...
+          <Tab eventKey={eventKey} title={`${tableName} (${count})`} disabled={isDisabled}>
+            {innerContents}
+          </Tab>
+        );
+      });
+    }
+    if (tables && tables.length > 0) {
+      const innerContents = activeTableTabKey === 'All' ? results : '';
+      tables.unshift(
+        <Tab eventKey="All" title="All">
```
luigivaldez, line 157: I might use a different (more complex/unusual) value for the key here, on the off chance that one of the bucket's is called "All"

src/components/SearchResults.js
```diff
@@ -91,6 +97,94 @@ export default class SearchResults extends React.Component<SearchResultsDefaultP
...
+          {innerContents}
+        </Tab>,
+      );
+      // `
```
luigivaldez, line 161: What's this comment for?

luigivaldez (changes requested):

src/components/FacetResults.js
brandon-bogan: Yes, exactly.

src/components/SearchResults.js
brandon-bogan: @luigivaldez I'm not sure what the pros and cons are of using the react-bootstrap components vs the SUIT components (I assume you're talking about TabPanel/TabInfo? Is the "standard" to use those over the bootstrap ones?

